PR #6, opened 2018-10-23
Support experimental Explore UI

ahayworth:
The Explore UI is in beta for Grafana 5.3.1. I thought it would be
nice if we could support arbitrary PNG rendering for any explore query.

We accomplish this by basically looking for a rendered canvas (because
the Explore UI doesn't emit the same events as a rendered dashboard).

Then, since there is no solo mode for explore panels, we render the
graph element instead.

### DIFF
--- a/src/browser.ts
+++ b/src/browser.ts
@@ -25,6 +25,7 @@ export class Browser {
   async render(options) {
     let browser;
     let page;
+    let exploreHandle;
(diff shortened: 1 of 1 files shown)


torkelo:
No, Sorry. rendering is only for dashboard panels.

torkelo:
You can open a feature request to support Explore as well so we can gauge interest in this feature.

PR #19, opened 2019-05-27
Package plugin via pkg PoC

aocenas:
Using [pkg](https://github.com/zeit/pkg) to create single binary with node baked in. There are few things that do not work right now:
- packaging puppeteer
- packaging grpc module which is native

This means both have to be included as a separate file/dir in the plugin. This results in package like this:
Screenshot 2019-05-27 at 17 31 51

That is then zipped and can be just unzipped in the grafana plugins dir.

TODO:
- [x] building for different architecture
- [x] puppeteer size, for some reason the version downloaded by puppeteer npm package is much larger than mac chromium app resulting in big ~600mb package (250 zipped) (was an issue with zip flattening symlinks)
- [x] cleanup hardcoded stuff like paths, arch names and so on

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,9 +16,8 @@ package:
 	node scripts/pkg.js ${ARCH}
 	node scripts/download_chromium.js ${ARCH}
 	node scripts/download_grpc.js ${ARCH}
+	node scripts/rename_executable.js ${ARCH}
 	cp plugin.json plugin-${ARCH}/
-	cp plugin-${ARCH}/renderer plugin-${ARCH}/plugin_start_linux_amd64
-	mv plugin-${ARCH}/renderer plugin-${ARCH}/plugin_start_darwin_amd64
 	tar -czf plugin-${ARCH}.tar.gz plugin-${ARCH}
 
 build_package: clean clean_package build package
--- a/Makefile
+++ b/Makefile
@@ -8,11 +8,10 @@ build:
 
 package:
 	rm -rf ./plugin
-	rm ./plugin.zip
+	rm -f ./plugin.tar.gz
 	./node_modules/.bin/pkg -t node10 . --out-path plugin
-	mkdir -p plugin/puppeteer
-	cp -R ./node_modules/puppeteer/.local-chromium plugin/puppeteer
-	cp node_modules/grpc/src/node/extension_binary/node-v64-darwin-x64-unknown/grpc_node.node plugin/
+	node download_chromium.js mac
+	node download_grpc.js darwin-x64-unknown
 	cp plugin.json plugin/
 	mv plugin/renderer plugin/plugin_start_darwin_amd64
-	zip -rq plugin plugin
+	tar -czf plugin.tar.gz plugin
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,6 @@ package:
 	node scripts/download_grpc.js ${ARCH}
 	node scripts/rename_executable.js ${ARCH}
 	cp plugin.json plugin-${ARCH}/
-	tar -czf plugin-${ARCH}.tar.gz plugin-${ARCH}
+	zip -yqr plugin-${ARCH}.zip plugin-${ARCH}
 
 build_package: clean clean_package build package
--- a/Makefile
+++ b/Makefile
@@ -5,3 +5,14 @@ clean:
 
 build:
 	tsc
+
+package:
+	rm -rf ./plugin
+	rm ./plugin.zip
+	./node_modules/.bin/pkg -t node10 . --out-path plugin
+	mkdir -p plugin/puppeteer
+	cp -R ./node_modules/puppeteer/.local-chromium plugin/puppeteer
+	cp node_modules/grpc/src/node/extension_binary/node-v64-darwin-x64-unknown/grpc_node.node plugin/
+	cp plugin.json plugin/
+	mv plugin/renderer plugin/plugin_start_darwin_amd64
+	zip -rq plugin plugin
--- a/Makefile
+++ b/Makefile
@@ -1,17 +1,24 @@
+ARCH = darwin-x64-unknown
+
 all: clean build
 
 clean:
-	rm -rf dist
+	rm -rf build
 
 build:
-	tsc
+	./node_modules/.bin/tsc
+
+clean_package:
+	rm -rf ./plugin-${ARCH}
+	rm -f ./plugin-${ARCH}.tar.gz
 
 package:
-	rm -rf ./plugin
-	rm -f ./plugin.tar.gz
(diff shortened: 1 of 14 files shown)


src/browser.ts
```diff
@@ -33,7 +33,12 @@ export class Browser {
       // set env timezone
       env.TZ = options.timezone || process.env.TZ;
 
-      browser = await puppeteer.launch({env: env, args: ['--no-sandbox']});
+      const executablePath = puppeteer.executablePath().replace(/^.*?\/node_modules/, path.dirname(process.execPath));
```
aocenas: As puppeteer isn't baked in the path needs to be changed here. There is `process.pkg` property that can be checked so this still works outside of the binary.

Makefile
```diff
@@ -5,3 +5,14 @@ clean:
 
 build:
 	tsc
+
+package:
+	rm -rf ./plugin
+	rm ./plugin.zip
+	./node_modules/.bin/pkg -t node10 . --out-path plugin
```
aocenas: Seems like node10 is the newest one they have.

aocenas:
So far I did not find anything that has more features than pkg so seems like packaging native code is an issue. The question is whether we really want this to be single binary or are ok with having a simple zip that can be just unzipped into plugins dir without any more setup.

aocenas:
Other things I tried so far:
- https://github.com/pmq20/node-packer
Does not work with puppeteer bc/ `Failed to launch chrome!`. Compilation time is long (like 26m) seems like it is compiling node from scratch. Compiler needs quite a bit of dependencies
- https://github.com/NectarJS/nectarjs
Got compilation errors that I did not solve. Little documentation, needs id and key to work, so not even sure if it is a webservice or what.
- https://github.com/nexe/nexe
Only looked at docs. From that it has the same shortcomings as pkg (cannot package native code into the binary), but less stars.

aocenas:
Added a script to download both chrome and grpc prebuild binaries. Those should be some what future proof and should work with future versions of both puppeteer and grpc-node packages. At the moment that means this can build the plugin for multiple platforms/archs.

torkelo:
that great! progress :)

aocenas:
When using grafana docker image there are quite a few things that need to be added for chrome to work:
```
RUN apt-get update && apt-get -y upgrade && apt-get install -y libasound2 libx11-6 libx11-xcb1 libxcomposite1 \
libxcursor1 libxdamage1 libxext6 libxi6 libxtst6 libglib2.0-0 libnss3 libcups2 libxss1 libxrandr2 libpangocairo-1.0-0 \
libatk1.0-0 libatk-bridge2.0-0 libgtk-3-0
```

aocenas:
Tested in linux (our docker image with mentioned caveats), mac and win (virtual box). There is an issue in latest pkg version with cross compiling (https://github.com/zeit/pkg/issues/688) but 4.3.8 works fine.

aocenas:
@torkelo Yes the thinking right now (@briangann can correct me here if I have something wrong), is that we can leave creating and hosting the packages on 3rd party mainly github (but probably anything will work).

So when you install plugin through our cli it goes to our API with plugin metadata, where we need to add something like `available_architectures` attribute that will tell you what architectures the plugin is available for and a url for version and arch combination. Then we can either download it by proxying through our API or just download directly from GH, and then unpack to plugin dir as we do now.

Looking at the plugin download code right now I think it should not be much work, we already list versions as commit hash and download stuff from GH if I am correct so this would just change that to either literal url (that way it can be stored anywhere, not sure if we want that), or just a repo url and then we will require specific naming for the releases (probably just <platform>_<arch>.tar.gz or something like that) so we can stitch the url together automatically.

Later on we can still add some CI and build and store the binaries ourselves but that looks like a much more complex task and not sure if it is must have right now.

torkelo:
That sounds like a good plan and exactly what I had in mind.

Yea, agree the CI part can wait.

torkelo:
Is there anything left you want to do here before merge?

aocenas:
@torkelo I would like to wait for https://github.com/raintank/grafana-net/pull/790 and https://github.com/grafana/grafana/pull/17551 when that is final I assume we can merge this, create a release and add it to the plugins repo and remove phantom js in grafana. As I have a 2 week vacation starting next week I am not sure I will be able to finish that for 6.3.

marefr:
> Rendering times out:
>
>> EROR[08-13|07:20:28] Rendering failed. logger=context userId=1 orgId=1 uname=admin error="Rendering failed: TimeoutError: waiting for function failed: timeout 60000ms exceeded"
EROR[08-13|07:20:28] Request Completed logger=context userId=1 orgId=1 uname=admin method=GET path=/render/dashboard-solo/new status=500 remote_addr=127.0.0.1 time_ms=62004 size=1722 referer="http://localhost:3001/dashboard/new?panelId=2&edit&fullscreen&orgId=1"

Need to save dashboard before being able to render

papagian:
> Need to save dashboard before being able to render

That worked, my mistake.

aocenas:
@papagian Thanks for testing, I will still check the issues you have on the ubuntu vm. The building should work after `yarn install` there shouldn't be anything else needed for this to work. Also the shared libraries I thought are only needed on stripped down docker containers.

papagian:
> @papagian Thanks for testing, I will still check the issues you have on the ubuntu vm. The building should work after `yarn install` there shouldn't be anything else needed for this to work. Also the shared libraries I thought are only needed on stripped down docker containers.

The `node` version was really old. After upgrading and running:
`yarn install --pure-lockfile` the building works fine.

papagian:
Additionally I have built it and tested successfully in ubuntu 18.04 and redhat 8.
For the latter, I had to run:
`yum -y install libX11 libXcomposite libXcursor libXdamage libXext libXi libXtst cups-libs libXScrnSaver libXrandr alsa-lib pango atk at-spi2-atk gtk3 `
following this [recommendation](https://github.com/GoogleChrome/puppeteer/issues/391#issuecomment-398660894)

marefr:
@aocenas I would like to merge this so we can look into publishing this on grafana.com and start getting usage and feedback of this. The thing you wanted to look into, is that related to building on ubuntu or running a built executable on ubuntu?

aocenas:
@marefr yes I just wanted to make sure it is clear when you need all the additional linux libs install. I mean probably not a blocker to just merge this and publish silently right now. I just hope to figure it out so at least on Ubuntu it is can be zero dependency install.

aocenas:
@marefr did few quick changes
- changed packaging to zip as that is what grafana cli will expect
- added readme about the packaging command

Also tested it in virtualbox with latest Ubunut LTS version and did not need any extra things to install so not sure where the difference is with @papagian tests

PR #23, opened 2019-08-16
Upgrade dependencies

marefr:
Ran `yarn upgrade`. Not sure if that's the best way to do this and/or this can break things :smile: 

Should solve repository security vulnerabilities reported in our dependencies.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -35,42 +35,42 @@
   dependencies:
     "@types/node" "*"
 
-"@types/events@*":
-  version "1.2.0"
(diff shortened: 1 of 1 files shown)


aocenas:
Well I think this should be fine should upgrade only things according to semver specifier in package.json, on the other hand it updates pupeteer from 1.7.x - 1.19.x so should be probably tested. No automated tests yet :(. Did you or do you have time to test it? Otherwise I could but not sure if I manage today.

marefr:
I only tested with docker and that worked. Probably need to test with plugin as well I suppose?

marefr:
@aocenas tried to only upgrade "needed" packages, but then I couldn't get the renderer to build/render.

Issue #16, opened 2019-05-07
Rendering not waiting for panel to load


mbudano:
Is there a way to set a time to wait for the panel to load? I'm having problems with several dashboards that are been render before the panels in them are ready (with data).
So I end up with an image of several panels been loaded.

hailong1004:
Have the same problem, set different timeout, but it does not take effect, when the panel needs more than 10s to get the data, those pictures have no data, my grafana version is 6.2.5

kjetilmjos:
Did you find out if there is some solution to the rendering time issue?

MarkTurbo:
I'm hitting exactly the same issue. Getting back PNGs of panels with the spinning 'loading' icon in the corner! Any updates about this issue?

kjetilmjos:
I did not find anything more about it

mbudano:
I was unable to find a solution, I was using the render function to show an image in a nodejs program, but eventually I changed the code to use puppeteer library

marefr:
@MarkTurbo are you rendering a single panel or a full dashboard?

MarkTurbo:
@marefr A full dashboard that consists of 15 SingleStat panels...

![image](https://user-images.githubusercontent.com/29234864/63187372-92201b00-c013-11e9-87f4-c2b459022064.png)

marefr:
Okay, thanks. This is currently not built for supporting rendering full dashboards, but probably something we would like to support in the future.

NateZimmer:
A query parameter for waiting would be useful.. I keep getting blank tables: 
![image](https://user-images.githubusercontent.com/1478536/80246524-2b3b1300-8632-11ea-9a2a-3e8d3598c5a8.png)

marefr:
@NateZimmer is that the buildin table panel or a community plugin?

NateZimmer:
@marefr that is the built in table panel. Using Grafana v6.3.5. I have this occasionally happen on plots as well. 

My temporary hack-a-round is to set an image size threshold and if the downloaded image is below the anticipated size, re-download.

alitoufighi:
Experiencing the same issue when setting a long time range. When I use 2~3 days, it renders fine, but for the range of 7 days it fails and I only get a "No Data". The panels are just normal graphs, reading data from a Prometheus datasource (configured to read metrics in the server-side).

alitoufighi:
[...]

PR #24, opened 2019-08-25
Wait until all network connections to be idle before rendering

d1ff:
Consider using waitUntil in page.goto to fix empty pngs

Sometimes 'load' event fires before query data is fetched, and page.waitForFunction bit doesn't help with that. This fix uses puppeteer's 'networkidle0' event, which ensures there is no active network requests for 500ms.

### DIFF
--- a/src/browser.ts
+++ b/src/browser.ts
@@ -66,7 +66,8 @@ export class Browser {
         'domain': options.domain,
       });
 
-      await page.goto(options.url);
+      // wait until all data was loaded
+      await page.goto(options.url, { waitUntil: 'networkidle0' });
 
       // wait for all panels to render
       await page.waitForFunction(() => {


marefr:
@d1ff thanks. Are there any issue opened that this PR solves?

d1ff:
#16 is very similar to mine, I've ran into this problem with table panels. Sadly, I do not have test dashboard or data to share.

kjetilmjos:
I just tested this PR and it fixes #16

kjetilmjos:
Tested with `networkidle2` and that resulted in panels being loaded without data again. This only happens if the time period selected is large and there is a lot of data. `networkidle0` seems to be working better as far as I can tell.

marefr:
Understood. Need to think some more about this and what possible edge cases may break by waiting for all network activity to be idle.

marefr:
@d1ff thank you

PR #40, opened 2019-10-02
Service: Support configuring host and port thru config and environment variables

kaey:
This allows users to run server listening on localhost.


marefr:
@kaey Thank you for contributing!

PR #51, opened 2019-11-13
Update Dockerfile to fix chromium segfault

mattes:
Running Chrome Headless inside a Docker container requires some more work. Currently none of the grafana-image-renderer images start for me.

```
docker run -it --rm grafana/grafana-image-renderer
Segmentation fault (core dumped)
```

Running with `--privileged` fixes it, but it's not a great solution.

```
docker run -it --rm --privileged grafana/grafana-image-renderer
HTTP Server started, listening on 8081
```

[Zenika/alpine-chrome](https://github.com/Zenika/alpine-chrome#3-ways-to-use-chrome-headless-with-this-image) mentions there is three ways to run chrome headless inside docker:

* With --no-sandbox
* With SYS_ADMIN capability
* With seccomp

But according to [Headless Chrome FAQ](https://developers.google.com/web/updates/2017/04/headless-chrome#faq):
> --no-sandbox is not needed if you properly setup a user in the container.

---

__This PR updates the Dockerfile to set up and use a `chrome` user inside the container so that the container can be started without any further security flags.__

Some more notes:
* I tried to stick with alpine first, but even after adding a user and group with (adduser/addgroup) it would still segfault.
* Buster is necessary for `fonts-open-sans`.
* I removed `udev`, I don't think it's needed.
* `ca-certificates` might actually not be needed, but it probably doesn't hurt to install.
* ENV vars are set in the correct build step now.
* The image is now 761 MB vs 536 MB before.
* I used  [lighthousebot's Dockerfile](https://github.com/GoogleChromeLabs/lighthousebot/blob/master/builder/Dockerfile) as a reference.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,40 +1,44 @@
-FROM node:10-alpine AS base
+FROM node:10-buster-slim AS base
 
-ENV CHROME_BIN="/usr/bin/chromium-browser"
+RUN apt-get update \
+    && apt-get install -y chromium fonts-open-sans ca-certificates --no-install-recommends \
+    && rm -rf /var/lib/apt/lists/* \
+    && rm -rf /src/*.deb
+
+ADD https://github.com/Yelp/dumb-init/releases/download/v1.2.2/dumb-init_1.2.2_amd64 /usr/local/bin/dumb-init
+RUN chmod +x /usr/local/bin/dumb-init
+
+
+FROM base as build
+
 ENV PUPPETEER_SKIP_CHROMIUM_DOWNLOAD="true"
 
 WORKDIR /usr/src/app
 
-RUN \
-  echo "http://dl-cdn.alpinelinux.org/alpine/edge/community" >> /etc/apk/repositories \
-  && echo "http://dl-cdn.alpinelinux.org/alpine/edge/main" >> /etc/apk/repositories \
-  && echo "http://dl-cdn.alpinelinux.org/alpine/edge/testing" >> /etc/apk/repositories \
-  && apk --no-cache  update \
-  && apk --no-cache  upgrade \
-  && apk add --no-cache --virtual .build-deps \
-    udev ttf-opensans chromium \
-		ca-certificates dumb-init \
-  && rm -rf /var/cache/apk/* /tmp/*
-
-FROM base as build
-
-COPY . ./
+COPY . .
 
 RUN yarn install --pure-lockfile
 RUN yarn run build
 
-EXPOSE 8081
-
-CMD [ "yarn", "run", "dev" ]
 
 FROM base
+
+ENV CHROME_BIN="/usr/bin/chromium"
 
 COPY --from=build /usr/src/app/node_modules node_modules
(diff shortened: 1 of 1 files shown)


marefr:
Thanks. No other similar reports like this. What architecture, operating system etc are you using?

mattes:
```
$ uname -a
Linux instance-1 5.0.0-1025-gcp #26~18.04.1-Ubuntu SMP Mon Nov 11 13:09:18 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux

$ docker version
Client: Docker Engine - Community
 Version:           19.03.4
 API version:       1.40
 Go version:        go1.12.10
 Git commit:        9013bf583a
 Built:             Fri Oct 18 15:54:09 2019
 OS/Arch:           linux/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          19.03.4
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.12.10
  Git commit:       9013bf583a
  Built:            Fri Oct 18 15:52:40 2019
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.2.10
  GitCommit:        b34a5c8af56e510852c35414db4c1f4fa6172339
 runc:
  Version:          1.0.0-rc8+dev
  GitCommit:        3e425f80a8c931f88e6d94a8c831b9d5aa481657
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683
```

This is essentially a plain Google Cloud Instance with Ubuntu Bionic running it. In case you'd like to reproduce. But yeah, I was surprised that nobody reported something before me.

LindezaGrey:
hey @mattes ,
thanks for your investigative work! I encountered the same issue but only on my cloud machines, not on my local dev machine. I received also segfault errors which i could not explain. My "solution" was to switch to a non alpine base image (node:10 instead of node:10-alpine). But this was more of a "try until it works" solution... i might try if your solution works for me as well!

hekmon:
Not 100% sure if related but this is the first lead I found for the issue I am having: on AWS ECS (w Fargate) the container does not start: it exists with 139 (segfault) and without any logs (captured by ECS). I tried with 1.0.5, 1.0.6 & 1.0.7.

hekmon:
We just tested @mattes Dockerfile on ECS: it works fine :)

aknuds1:
@marefr Should we try to test this to see if we can reproduce the error?

marefr:
@aknuds1 would wait some more. Still very few reported this problem.

hekmon:
@marefr you just have GCP & AWS not working ;)

You should definitely test it on your side.

On our side the env is the following:
AWS ECS/Fargate 1.3.0 (latest)

mattes:
Related #70

hekmon:
Thx @mattes ! I can confirm that v1.0.8-beta2 works on AWS ECS Fargate 1.3.0.

marefr:
Thank you. Will release v1.0.8 stable soon so closing this.

PR #53, opened 2019-11-18
Docker: Updates dependencies to remove vulnerabilities

marefr:
**Before update:**
```bash
> docker run --rm -v $HOME/.cache:/root/.cache/ aquasec/trivy --clear-cache grafana/grafana-image-renderer:latest                                                                                                                             
2019-11-18T22:24:46.663Z        INFO    Removing image caches...                                                                                                                                                                              
2019-11-18T22:24:46.668Z        INFO    Updating vulnerability database...                                                                                                                                                                    
2019-11-18T22:25:04.827Z        INFO    Detecting Alpine vulnerabilities...                                                                                                                                                                   
2019-11-18T22:25:04.834Z        INFO    Updating yarn Security DB...                                                                                                                                                                          
2019-11-18T22:25:06.067Z        INFO    Detecting yarn vulnerabilities...                                                                                                                                                                     

grafana/grafana-image-renderer:latest (alpine 3.9.4)
====================================================
Total: 12 (UNKNOWN: 0, LOW: 2, MEDIUM: 6, HIGH: 4, CRITICAL: 0)

+----------+------------------+----------+-------------------+---------------+--------------------------------+
| LIBRARY  | VULNERABILITY ID | SEVERITY | INSTALLED VERSION | FIXED VERSION |             TITLE              |
+----------+------------------+----------+-------------------+---------------+--------------------------------+
| curl     | CVE-2019-5482    | HIGH     | 7.64.0-r2         | 7.64.0-r3     | curl: heap buffer overflow in  |
|          |                  |          |                   |               | function tftp_receive_packet() |
+          +------------------+          +                   +               +--------------------------------+
|          | CVE-2019-5481    |          |                   |               | curl: double free due to       |
|          |                  |          |                   |               | subsequent call of realloc()   |
+----------+------------------+----------+-------------------+---------------+--------------------------------+
| expat    | CVE-2019-15903   | MEDIUM   | 2.2.7-r0          | 2.2.7-r1      | expat: heap-based buffer       |
|          |                  |          |                   |               | over-read via crafted XML      |
|          |                  |          |                   |               | input                          |
+----------+------------------+          +-------------------+---------------+--------------------------------+
| libxslt  | CVE-2019-18197   |          | 1.1.33-r1         | 1.1.33-r2     | libxslt: use after             |
|          |                  |          |                   |               | free in xsltCopyText in        |
|          |                  |          |                   |               | transform.c leads to password  |
|          |                  |          |                   |               | disclosure...                  |
[...]

PR #55, opened 2019-11-21
Docker: Fix Dockerfile

marefr:
Based on https://github.com/grafana/grafana-image-renderer/pull/53#issuecomment-556344273 cleaning up the Dockerfile. Note: The only way I could properly remove nghttp2 openldap curl packages was to first install them, not sure if there are any easier/better way?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,6 @@ RUN \
   echo "http://dl-cdn.alpinelinux.org/alpine/edge/community" >> /etc/apk/repositories && \
   echo "http://dl-cdn.alpinelinux.org/alpine/edge/main" >> /etc/apk/repositories && \
   echo "http://dl-cdn.alpinelinux.org/alpine/edge/testing" >> /etc/apk/repositories && \
-  apk --no-cache update && \
   apk --no-cache upgrade && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -6,15 +6,14 @@ ENV PUPPETEER_SKIP_CHROMIUM_DOWNLOAD="true"
...
-  && apk --no-cache  update \
-  && apk --no-cache  upgrade \
-  && apk add --no-cache --virtual .build-deps \
-    udev ttf-opensans chromium \
-		ca-certificates dumb-init curl nghttp2 openldap \
-  && rm -rf /var/cache/apk/* /tmp/*
+  echo "http://dl-cdn.alpinelinux.org/alpine/edge/community" >> /etc/apk/repositories && \
+  echo "http://dl-cdn.alpinelinux.org/alpine/edge/main" >> /etc/apk/repositories && \
+  echo "http://dl-cdn.alpinelinux.org/alpine/edge/testing" >> /etc/apk/repositories && \
+  apk --no-cache update && \
```
aknuds1: Could you remove this command, since it has no effect (to my knowledge)?

PR #57, opened 2019-11-24
Upgrade Node to LTS (12)

marefr:
Upgrades Node from version 10 to LTS, version 12.


src/grpc-plugin.ts
```diff
@@ -1,24 +1,29 @@
...
+
+export const rendererProtoDescriptor = grpc.loadPackageDefinition(renderPackageDef);
 
 export class GrpcPlugin {
   constructor(private log: Logger, private browser: Browser) {}
 
   start() {
     var server = new grpc.Server();
 
-    server.addService(GRPC_HEALTH_PROTO.Health.service, {
```
aknuds1: Are you supposed to remove this health service?

Makefile
```diff
@@ -9,11 +9,11 @@ clean:
 
 deps: node_modules
 
-node_modules: package.json yarn.lock
+node_modules: yarn.lock
 	@echo "install frontend dependencies"
 	yarn install --pure-lockfile --no-progress
 
-build:
+build: clean
```
aknuds1: Why is it necessary to clean before building? Just asking since I don't know :)

src/grpc-plugin.ts
marefr: Upgraded grpc and got couldn't add the health service any longer. Read [here](https://grpc.io/docs/tutorials/basic/node/) and nothing mentioned the health stuff so tried to just remove it and seems to work :)

Makefile
marefr: Good question. Couldn't build if the build directory existed, but now looking at it think "phony" maybe are needed to not confuse make with command and directories/files?

src/grpc-plugin.ts
marefr: Added it back. Apparently the hashicorp/go-plugin requires it: https://github.com/hashicorp/go-plugin/blob/master/docs/guide-plugin-write-non-go.md#3-add-the-grpc-health-checking-service

However, only seems to be called when client calls [Ping](https://github.com/hashicorp/go-plugin/blob/e8d22c780116115ae5624720c9af0c97afe4f551/protocol.go#L44) which Grafana currently doesn't do.

Makefile
marefr: Added phony and cleaned up the makefile.

src/grpc-plugin.ts
DanCech: Interesting, for running in k8s it would also be nice to have a /healthz http endpoint, right now I've been using / for readiness checks.

PR #80, opened 2020-01-09
Support of PDF

bgranvea:
This PR extends rendering with support of PDF. The parameter 'encoding' must be set to 'pdf'.

There is a new 'jsonData' parameter which allows to inject extra parameters at different steps of the rendering process (see https://github.com/grafana/grafana/pull/21307 for support of this parameter in Grafana). 

'jsonData' is a JSON object with the following possible fields:

- launchOptions: is sent as parameter of launch method (see https://github.com/puppeteer/puppeteer/blob/master/docs/api.md#puppeteerlaunchoptions)

Example:
`{"launchOptions":{"headless":false},"waitFor":30000}`
(can be useful to debug headless browser)

- viewport: is sent as parameter of setViewport method (see https://github.com/puppeteer/puppeteer/blob/master/docs/api.md#pagesetviewportviewport).

Example:
`{"viewport":{"deviceScaleFactor":4}}`

- emulateMedia: parameter of https://github.com/puppeteer/puppeteer/blob/master/docs/api.md#pageemulatemediatype

- defaultNavigationTimeout: parameter of https://github.com/puppeteer/puppeteer/blob/master/docs/api.md#pagesetdefaultnavigationtimeouttimeout

- extraUrlParams: this string is added at the end of the Grafana URL which is called.

Example: 
`{"extraUrlParams": "&autofitpanels"}`

- styleTags: an array of strings which are passed to addStyleTag method 
https://github.com/puppeteer/puppeteer/blob/master/docs/api.md#pageaddstyletagoptions

Example:
`{"styleTags":[{"content":".sidemenu {display: none;}"}]}`
(hide left menu)

- scriptTags: an array of strings which are passed to addScriptTag method https://github.com/puppeteer/puppeteer/blob/master/docs/api.md#pageaddscripttagoptions

This could allow to inject specific Javascript but I have no use case.

- waitFor: number of milliseconds to wait after loading the page

- pdf: options passed to https://github.com/puppeteer/puppeteer/blob/master/docs/api.md#pagepdfoptions

Example:
`{"pdf":{"format":"A4","printBackground":true,"scale":0.6}}`

- screenshot: options passed to https://github.com/puppeteer/puppeteer/blob/master/docs/api.md#pagescreenshotoptions

Example:
`{"screenshot":{"fullPage":true}}`
(is used only for PNG/JPEG rendering)

### DIFF
--- a/src/browser.ts
+++ b/src/browser.ts
@@ -1,8 +1,11 @@
 import * as os from 'os';
 import * as puppeteer from 'puppeteer';
+import * as boom from 'boom';
 import { Logger } from './logger';
 import uniqueFilename = require('unique-filename');
 import { RenderingConfig } from './config';
+
+const allowedFormats: string[] = ['png', 'jpeg', 'pdf'];
 
 export class Browser {
   chromeBin?: string;
@@ -22,6 +25,20 @@ export class Browser {
     if (options.height > 3000 || options.height < 10) {
       options.height = 1500;
     }
+
+    if (options.encoding === '') {
+      options.encoding = 'png';
+    }
+
+    if (allowedFormats.indexOf(options.encoding) === -1) {
+      throw boom.badRequest('Unsupported encoding ' + options.encoding);
+    }
+
+    if (options.jsonData) {
+      options.jsonData = JSON.parse(options.jsonData);
+    } else {
+      options.jsonData = {};
+    }
   }
 
   async render(options) {
@@ -32,6 +49,8 @@ export class Browser {
     try {
       this.validateOptions(options);
 
+      this.log.debug('JSON data: %j', options.jsonData);
+
       // set env timezone
       env.TZ = options.timezone || process.env.TZ;
 
@@ -39,6 +58,7 @@ export class Browser {
         env: env,
         ignoreHTTPSErrors: this.config.ignoresHttpsErrors,
         args: ['--no-sandbox'],
+        ...options.jsonData.launchOptions,
       };
 
       if (this.config.chromeBin) {
@@ -51,8 +71,16 @@ export class Browser {
       await page.setViewport({
         width: options.width,
         height: options.height,
-        deviceScaleFactor: 1,
+        ...options.jsonData.viewport,
       });
+
+      if (options.jsonData.emulateMedia) {
+        await page.emulateMedia(options.jsonData.emulateMedia);
+      }
+
+      if (options.jsonData.defaultNavigationTimeout) {
+        await page.setDefaultNavigationTimeout(options.jsonData.defaultNavigationTimeout);
+      }
 
       await page.setCookie({
         name: 'renderKey',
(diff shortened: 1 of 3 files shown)


mr-yaky:
@bgranvea thank you for the good PR. I have tested it and it works nice! I have one question: did you tried to print some long pages as grafana entire dashboard ? I have played a bit but always get single page for PDF. Checked the code of plugin but didn't find any specific limitations for this staff. Also I tried to increase the `height` for `viewPort` but without success.   

Do you have some ideas why it not generates multi-pages PDF docs ? I used your code completely from your fork based on `1.0.8-beta2` version.

mr-yaky:
don't know a lot about this `puppeteer` library, so may be I missed something.

rohanprasadofficial:
@mr-yaky @bgranvea  Is there any way to generate pdf from png we get, I used some other libraries but while return response I don't see anything on the screen. 
Can you help me?  
Thanks

bgranvea:
> @bgranvea thank you for the good PR. I have tested it and it works nice! I have one question: did you tried to print some long pages as grafana entire dashboard ? I have played a bit but always get single page for PDF. Checked the code of plugin but didn't find any specific limitations for this staff. Also I tried to increase the `height` for `viewPort` but without success.
> 
> Do you have some ideas why it not generates multi-pages PDF docs ? I used your code completely from your fork based on `1.0.8-beta2` version.

you have to use some CSS hack as explained here: https://community.grafana.com/t/dashboard-printing-cannot-scroll-past-first-page/9726

it can be injected in the page through scriptTags parameter:

`{"styleTags":[{"content":".scroll-canvas {all: unset;}"}]}
`

Note however that page breaks cannot be controlled, so in my use case, I try to stay with a single page PDF.

rohanprasadofficial:
@bgranvea  so my issues is wherever I am hitting  http://localhost:3500/render/dashboard/db/home?orgId=1&width=2000&height=2500  , I am getting a png return and I know the path of it (results.filePath) but now I have used npm module to convert it to pdf but now when I do res.sendFile() it shows blank screen .
I have the master branch code , I haven't incorporate you PR yet

mr-yaky:
@bgranvea thank you for the recommendation. Yes, yesterday I tough about the same way with using `page.addStyleTag(options)` because I saw that even with printing any dashboard as PDF directly in Chromium I also got one page. 

I have checked with :
```
{"styleTags":[{"content":".scroll-canvas {all: unset;}"}]}
```

and I always get an error with parsing of `jsonData`:

```
SyntaxError: Unexpected end of JSON input
    at JSON.parse (<anonymous>)
    at Browser.validateOptions (/usr/src/app/build/browser.js:41:37)
    at Browser.<anonymous> (/usr/src/app/build/browser.js:53:22)
    at Generator.next (<anonymous>)
    at /usr/src/app/build/browser.js:7:71
    at new Promise (<anonymous>)
    at __awaiter (/usr/src/app/build/browser.js:3:12)
    at Browser.render (/usr/src/app/build/browser.js:48:16)
    at HttpServer.<anonymous> (/usr/src/app/build/service/http-server.js:37:47)
    at Generator.next (<anonymous>) {
  isBoom: true,
  isServer: true,
  data: null,
  output: {
    statusCode: 500,
    payload: {
      statusCode: 500,
      error: 'Internal Server Error',
      message: 'An internal server error occurred'
    },
    headers: {}
  },
  reformat: [Function],
  isDeveloperError: true
}
```

If I remove the `;` symbol from style fix then it allow the query. I tried different ways but still not solve this issue. Seems that parser understands that as Json part.

mr-yaky:
Finally, I have fixed the issue with hardcoding needed styles injection directly to plugin code. I found that problem was in a parsing while sending as jsonData.

mr-yaky:
@rohanprasadofficial don't know because I have tried only with creating PDF on the fly with PRs from @bgranvea

rohanprasadofficial:
@mr-yaky  does it gives the PDF on one page ? does it look exactly similar to the dashboard?

bgranvea:
> Finally, I have fixed the issue with hardcoding needed styles injection directly to plugin code. I found that problem was in a parsing while sending as jsonData.

it should work by putting it in jsonData argument, there must some encoding problem.

mr-yaky:
@bgranvea can you take a look when you will have a time, please ? It would be great to have this fix.

marefr:
We'll not accept this now. Please refer to https://github.com/grafana/grafana/pull/21307#issuecomment-610436075. Would suggest to close this since we're trying to keep PR's open that are relevant - or would you like to keep this open and continiue to work on it?

bgranvea:
ok, I close it.

PR #82, opened 2020-01-13
Add SKIP_GRPC build option to allow local builds

witekest:
The option allows using gprc from the local build instead of downloading
the precompiled binary.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,7 @@
 ARCH = darwin-x64-unknown
 SKIP_CHROMIUM =
 OUT =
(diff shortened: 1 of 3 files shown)


marefr:
Why doesn't the precompiled binary work?

witekest:
I was hoping to build the package without downloading anything but noticed it's not the end of problems. Eventually I've sticked with the [local build procedure](https://github.com/grafana/grafana-image-renderer/blob/master/docs/building_from_source.md#local-nodejs-application-using-local-chromechromium).

PR #83, opened 2020-01-14
Render image even if error alert appears

sdwire:
Prior to this PR, any error in a grafana panel will cause the waitFunction to never return, causing a timeout error. This makes debugging grafana very difficult.

This PR allows errors to be visible in the rendered screenshot

### DIFF
--- a/src/browser.ts
+++ b/src/browser.ts
@@ -63,11 +63,11 @@ export class Browser {
       // wait until all data was loaded
       await page.goto(options.url, { waitUntil: 'networkidle0' });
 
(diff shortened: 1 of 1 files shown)


marefr:
Not sure this will work since alert may be shown even if panels render data. I've just merged #88 which should capture more debug information from browser/console and it will be included in v1.0.8 stable version.

marefr:
Thanks, but I'm closing this for now. A more appropriate solution, if needed, would be a setting to enable screenshot to be taken upon timeout

PR #94, opened 2020-01-30
Upgrade deps

marefr:
For example Puppeteer 2.0.0 => 2.1.0 which seems like a big change, see [release](https://github.com/puppeteer/puppeteer/releases/tag/v2.1.0), so we might want to be a bit careful with this one.


aknuds1:
It does look as if 2.1.0 shouldn't break anything unless we use Firefox with it already? Are you thinking the Chromium upgrade might break things?

marefr:
Superseded by recent upgrade

PR #110, opened 2020-02-27
Renderer: add device scale support

shankerwangmiao:
This pull request adds support for `scale` parameter to the renderer, through both the gRPC interface and the http interface.

For interoperability with the current version of Grafana, the missing  `scale` field will be treated as 1, so that the images can be rendered as before. 

ref: https://github.com/grafana/grafana/issues/22472

### DIFF
--- a/src/plugin/grpc-plugin.ts
+++ b/src/plugin/grpc-plugin.ts
@@ -79,7 +79,7 @@ export class GrpcPlugin {
       url: req.url,
       width: req.width,
       height: req.height,
-      scale: (req.scale > 0 && req.scale != Infinity) ? req.scale : 1,
+      scale: req.scale,
       filePath: req.filePath,
       timeout: req.timeout,
       renderKey: req.renderKey,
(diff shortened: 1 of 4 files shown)


src/plugin/grpc-plugin.ts
```diff
@@ -79,6 +79,7 @@ export class GrpcPlugin {
       url: req.url,
       width: req.width,
       height: req.height,
+      scale: (req.scale > 0 && req.scale != Infinity) ? req.scale : 1,
```
aknuds1: Might it be better to drop this validation, since we also validate later? If we were to change the default later f.ex., it might be bad to hardcode it twice. Bugs could also potentially arise from having two different validation logics.

shankerwangmiao: You are right about this.

marefr:
@shankerwangmiao thank you for your contribution. This have been replaced by #128 so will close this.

PR #114, opened 2020-03-11
Render: Fix panel titles should not be focused when rendering

### LINKED ISSUE #111: Image render shows popup description without disable popup option
**What happened**:
Images redered with Description bubble like on the image

**What you expected to happen**:
Rendered image without bubble (by default option) and ENV settings trigger like _GF_RENDER_DESCRIPTION_BUBBLES=true/false_
**How to reproduce it (as minimally and precisely as possible)**:
Setup description on any panel and render it
**Anything else we need to know?**:
We are use curl for renders without UI

**Environment**:
- Grafana Image Renderer version: 1.0.9 (docker)
- Grafana version: 6.5.3
- Installed plugin or remote renderer service: natel-discrete-pane , natel-influx-admin-panel, grafana-piechart-panel, vonage-status-panel, camptocamp-prometheus-alertmanager-datasource, grafana-clock-panel and remote render in docker
- OS Grafana Image Renderer is installed on: SLES 12,15, RHEL 7, OEL 7
- User OS & Browser: Win10, Latest Chrome
- Others: none

AgnesToulet:
Add a call to the mouse.move function so the mouse is in the bottom right corner of the viewport instead of the top left corner. 
It allows to not have the "mouse:hover" style on titles while rendering panels.

I've tested several calls to the render function and it seems not to have any other impact.

Fixes #111

### DIFF
--- a/src/browser/browser.ts
+++ b/src/browser/browser.ts
@@ -138,6 +138,7 @@ export class Browser {
       value: options.renderKey,
       domain: options.domain,
     });
+    await page.mouse.move(options.width, options.height);
 
     await this.timings.navigate(async () => {
       // wait until all data was loaded


marefr:
We just need to test it

aknuds1:
Any suggestions on how to test it @marefr? You know better than me how to look for eventual bugs from this change.

marefr:
@aknuds1 you can test referenced issue by creating a panel which have a panel link and use v1.0.10 to reproduce. Then checkout this branch and build a local plugin version or docker image and verify it resolves the issue. Please let me know if you would like me to do it.

PR #128, opened 2020-04-18
Adds support for Grafana v2 backend plugins

marefr:
Goes together with https://github.com/grafana/grafana/pull/23625

Implements backend image renderer plugin v2:
- Version negotiation enables correct plugin version to be run depending on what Grafana version.  V2 will be used if https://github.com/grafana/grafana/pull/23625 and V1 if master or older Grafana versions.
- Adds supports for setting viewport device scale factor in headless browser instance
- Adds support for sending Accept-Language HTTP headers in headless browser instance
- Adds support for parsing environment variables sent from Grafana
- V2 plugin implements Diagnostics gRPC service with health check and collect metrics

**Test/verification:**
- [x] Verified it works running plugin on Linux/Ubuntu with https://github.com/grafana/grafana/pull/23625 and Grafana master. 
Verified it works running as HTTP service on Linux/Ubuntu with https://github.com/grafana/grafana/pull/23625 and Grafana master. 
- [x] Verified plugin works on Windows
- [x] Verified plugin works on Mac


src/browser/browser.ts
```diff
@@ -64,13 +57,35 @@ export class Browser {
     options.height = parseInt(options.height as string, 10) || 500;
     options.timeout = parseInt(options.timeout as string, 10) || 30;
 
-    if (options.width > 3000 || options.width < 10) {
-      options.width = 2500;
+    if (options.width > this.config.maxWidth || options.width < 10) {
+      options.width = this.config.maxWidth;
+    }
+
+    if (options.width < 10) {
```
aknuds1, line 64: The only reason width could be less than 10 at this stage is if this.config.maxWidth is less than 10 right? Would it make sense to ensure this.config.maxWidth >= 10 earlier?

src/node-plugin/lib/server.ts
```diff
@@ -0,0 +1,131 @@
...
+  // set the old un-versioned fields as if they were versioned plugins
+  if (!opts.versionedPlugins) {
+    opts.versionedPlugins = {} as VersionedPluginSet;
+  }
+
+  if (pluginSet) {
+    opts.versionedPlugins[protoVersion] = pluginSet;
+  }
+
+  // Sort the version to make sure we match the latest first
```
aknuds1, line 50:
```suggestion
  // Sort the versions to make sure we match the latest first
```

src/plugin/v2/grpc_plugin.ts
```diff
@@ -0,0 +1,207 @@
...
+    let browserVersion = 'unknown';
+
+    try {
+      browserVersion = await browser.getBrowserVersion();
+      metrics.browserInfo.labels(browserVersion).set(1);
+      this.log.debug('Using browser version', 'browserVersion', browserVersion);
+    } catch {
+      this.log.error('Failed to get browser version');
+      metrics.browserInfo.labels('unknown').set(0);
+    }
```
aknuds1, line 65: As little as possible should be in a try block:
```suggestion
    let browserVersion = 'unknown';
    let labelValue = 1;

    try {
      browserVersion = await browser.getBrowserVersion();
    } catch (err) {
      this.log.error('Failed to get browser version', 'err', err);
      labelValue = 0;
    }
    metrics.browserInfo.labels(browserVersion).set(labelValue);
    if (browserVersion !== 'unknown') {
          this.log.debug('Using browser version', 'browserVersion', browserVersion);
    }
```

src/plugin/v2/grpc_plugin.ts
```diff
@@ -0,0 +1,207 @@
...
+    };
+
+    try {
+      this.log.debug('Render request received', 'url', options.url);
+      await this.browser.render(options);
+      callback(null, { error: '' });
+    } catch (err) {
+      this.log.error('Render request failed', 'url', options.url, 'error', err.toString());
+      callback(null, { error: err.toString() });
+    }
```
aknuds1, line 114:
```suggestion
    this.log.debug('Render request received', 'url', options.url);
    let errStr = '';
    try {
      await this.browser.render(options);
    } catch (err) {
      this.log.error('Render request failed', 'url', options.url, 'error', err.toString());
      errStr = err.toString();
    }
    callback(null, { error: errStr });
```

proto/pluginv2.proto
```diff
@@ -0,0 +1,171 @@
...
+  // Map of refId to response
+  map<string,DataResponse> responses = 1;
+}
+
+message DataResponse {
+  // Arrow encoded DataFrames
+  // Frame has its own meta, warnings, and repeats refId
+  repeated bytes frames = 1;
+  string error = 2;
+  bytes jsonMeta = 3; // Warning: Current ignored by frontend. Would be for metadata about the query.
```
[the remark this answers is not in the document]
marefr: Please suggest/make change in plugin sdk repo. This is a copy of that proto file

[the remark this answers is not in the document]
aknuds1: Aha, thanks, didn't realize that.

aknuds1:
Looks really good now, just wondering about my remaining comment :)

src/browser/browser.ts
marefr: Thanks for noticing. Now it uses the default width/height if width/height is less than 10.

PR #135, opened 2020-05-06
Plugin: Migrate to @grpc/grpc-js to resolve problems when IPv6 is disabled 

### LINKED ISSUE #48: Does not work when IPv6 is disabled
Is this plugin trying to bind to `"[::]"` ?

Because we disable IPv6 in the kernel to avoid a certain class of problems, and the plugin refuses to start with this message:
`E1023 17:01:02.999153762   20891 server_chttp2.cc:38]        {"created":"@1571842862.999100646","description":"No address added out of total 1 resolved","file":"../deps/grpc/src/core/ext/transport/chttp2/server/chttp2_server.cc","file_line":305,"referenced_errors":[{"created":"@1571842862.999093903","description":"Failed to add port to server","file":"../deps/grpc/src/core/lib/iomgr/tcp_server_uv.cc","file_line":436,"referenced_errors":[{"created":"@1571842862.999087065","description":"Failed to initialize UV tcp handle","file":"../deps/grpc/src/core/lib/iomgr/tcp_server_uv.cc","file_line":277,"os_error":"address family not supported"}]}]}`

aknuds1:
Migrate to @grpc/grpc-js, in order to solve #48, since the IPv6 issue is presumably not present in the pure-JS version of gRPC. 

I had to make a lot of changes because of TypeScript compilation errors that appeared, I can only guess that grpc-js has stricter type definitions. In particular, I had to move gRPC methods out of classes, because the class types didn't have the right interface. 

In retrospect, I think I might understand how to give the class types the right interface (basically make them indexable by string), so I could probably re-introduce classes for containing the methods.

Fixes #48

**NB:**
How should I best test this? I tested it manually, by rendering a panel in Grafana and it does work.


src/node-plugin/lib/server.ts
```diff
@@ -114,19 +113,26 @@ export const serve = async (opts: ServeConfig) => {
...
 
   opts.grpcHost = opts.grpcHost || '127.0.0.1';
   opts.grpcPort = opts.grpcPort || 0;
 
-  const address = `${opts.grpcHost}:${opts.grpcPort}`;
-  const boundPortNumber = await util.promisify(server.bindAsync).bind(server)(address, grpc.ServerCredentials.createInsecure());
-  if (boundPortNumber === 0) {
-    throw new Error(`failed to bind address=${address}, boundPortNumber=${boundPortNumber}`);
-  }
+  return new Promise<void>((resolve, reject) => {
```
aknuds1, line 123: Why is `new Promise` preferred to `util.promisify`? I think the latter is cleaner and more modern since you avoid mixing `async` and callback style code.

marefr: util.promisify didn't work with zeit/pkg that we're using to package plugin with node runtime.

marefr:
Works with Ubuntu 16.04 with both Grafana 6.7.x and 7.0.x. Still need to figure out how to disable IPv6 and test again.

src/node-plugin/lib/server.ts
aknuds1: Ah, that's too bad :/

marefr:
Tried disable IPv6 using these [instructions](https://www.neuraldump.net/2016/11/how-to-disable-ipv6-in-ubuntu-16-04-xenial-xerus/#comment-22453):
```
# create the long-life config file
echo "net.ipv6.conf.all.disable_ipv6 = 1
net.ipv6.conf.default.disable_ipv6 = 1
net.ipv6.conf.lo.disable_ipv6 = 1" | sudo tee /etc/sysctl.d/99-my-disable-ipv6.conf

# ask the system to use it
sudo service procps reload

# check the result
cat /proc/sys/net/ipv6/conf/all/disable_ipv6
```

Can still run the v1.0.12 version of image renderer plugin. Not sure what else to do?

marefr:
Was able to turn off IPv6 completely using a boot option ([instructions](https://askubuntu.com/a/337736))

- Couldn't start v1.0.12 on v6.7.x (should be due to IPv6 disabled).
- Can run this branch on v6.7.x and 7.0.x :+1:

aknuds1:
> Was able to turn off IPv6 completely using a boot option ([instructions](https://askubuntu.com/a/337736))
> 
> * Couldn't start v1.0.12 on v6.7.x (should be due to IPv6 disabled).
> * Can run this branch on v6.7.x and 7.0.x +1

@marefr Yayyy

marefr:
Since we're upgrading zeit/pkg need to verify these changes works on Windows and Darwin as well - had issues with that earlier.

marefr:
Verified works on Windows. Who can help out testing this on Darwin?

aknuds1:
@marefr I can do that.

aknuds1:
I tested it on OS X @marefr, works.

scripts/package_target.sh
```diff
@@ -25,11 +25,11 @@ if [ $? != 0 ]; then
...
 
-node scripts/download_grpc.js ${ARCH} ${OUT}
+# node scripts/download_grpc.js ${ARCH} ${OUT}
 
-if [ $? != 0 ]; then
-   echo "${?}\n". 1>&2 && exit 1
-fi
+# if [ $? != 0 ]; then
+#    echo "${?}\n". 1>&2 && exit 1
+# fi
```
marefr, line 32: Remove these comments

aknuds1: @marefr Done.

marefr: Ready to merge then :+1:

towolf:
Thanks @aknuds1! Great work.

PR #137, opened 2020-05-11
Sync latest changes of pluginv2 proto

marefr:
/api/plugins/grafana-image-renderer/health now works again.

### DIFF
--- a/proto/pluginv2.proto
+++ b/proto/pluginv2.proto
@@ -7,7 +7,13 @@ option go_package = ".;pluginv2";
 // Common
 //-----------------------------------------------
 
-message DataSourceConfig {
+message AppInstanceSettings {
+  bytes jsonData = 3;
+  map<string,string> decryptedSecureJsonData = 4;
+  int64 lastUpdatedMS = 5;
+}
+
+message DataSourceInstanceSettings {
   int64 id = 1;
   string name = 2;
   string url = 3;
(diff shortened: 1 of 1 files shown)


proto/pluginv2.proto
```diff
@@ -7,44 +7,59 @@ option go_package = ".;pluginv2";
...
-
 message User {
   string login = 1;
   string name = 2;
   string email = 3;
   string role = 4;
 }
 
+message PluginContext {
+  // The Grafana organization id the request originating from.
```
aknuds1, line 37:
```suggestion
  // The Grafana organization id the request originates from.
```

proto/pluginv2.proto
```diff
@@ -7,44 +7,59 @@ option go_package = ".;pluginv2";
...
+  int64 orgId = 1;
+
+  // The unique identifier of the plugin the request  originating from.
```
aknuds1, line 40:
```suggestion
  // The unique identifier of the plugin the request originates from.
```

proto/pluginv2.proto
```diff
@@ -7,44 +7,59 @@ option go_package = ".;pluginv2";
...
+  string pluginId = 2;
+
+  // The Grafana user the request originating from.
```
aknuds1, line 43:
```suggestion
  // The Grafana user the request originates from.
```

proto/pluginv2.proto
marefr: This should be changed in https://github.com/grafana/grafana-plugin-sdk-go/blob/master/proto/backend.proto

proto/pluginv2.proto
```diff
@@ -7,44 +7,59 @@ option go_package = ".;pluginv2";
...
+  //
+  // Will not be provided if Grafana backend initiated the request.
+  User user = 3;
+
+  // App plugin instance settings is the configured app instance settings.
+  // In Grafana an app instance is an enabled app plugin in a
+  // Grafana organization.
+  //
+  // Will only be set if request targeting an app instance.
```
aknuds1, line 52:
```suggestion
  // Will only be set if request is targeting an app instance.
```

proto/pluginv2.proto
```diff
@@ -7,44 +7,59 @@ option go_package = ".;pluginv2";
...
+  AppInstanceSettings appInstanceSettings = 4;
+
+  // Data source instance settings is the configured data source instance
+  // settings. In Grafana a data source instance is a created data source
+  // in a Grafana organization.
+  //
+  // Will only be set if request targeting a data source instance.
```
aknuds1, line 59:
```suggestion
  // Will only be set if request is targeting a data source instance.
```

proto/pluginv2.proto
aknuds1: @marefr OK thanks for the heads up.

marefr:
Yes, this is from sdk master

PR #144, opened 2020-05-19
Run image renderer under non-root Grafana user. Fixes #140

### LINKED ISSUE #140: Docker: Run image with non-root user
**What would you like to be added**:
Run docker image with non-root user.

**Why is this needed**:
Security

wardbekker:


VishalLahane:
@wardbekker  @aknuds1 why we use 472 id not more than 999 , will you plz explain ?

spaced:
PR is open since 2020. any news?

DanCech:
@VishalLahane 472 is just the https://en.wikipedia.org/wiki/ISO/IEC_9995#ISO/IEC_9995-8 numeric code for GRA ;)

VishalLahane:
@DanCech ideally it should be 1000+ as per https://medium.com/@gggauravgandhi/uid-user-identifier-and-gid-group-identifier-in-linux-121ea68bf510

PR #180, opened 2020-09-28
Allow configuring the default browser timeout

milesbxf:
This adds a new RENDERING_DEFAULT_TIMEOUT environment variable that sets the default page timeout for browser actions - see https://github.com/puppeteer/puppeteer/blob/main/docs/api.md#pagesetdefaulttimeouttimeout for what this covers.

We set the default to 30000, which is the Puppeteer default, so this won't change any existing behaviour unless the environment variable is explicitly set.


milesbxf:
I've since realised that this duplicates the functionality of the timeout parameter, so closing

PR #190, opened 2020-11-10
WIP: Chore: Switch from tslint to eslint

aknuds1:
Switch from tslint to eslint, since the former is deprecated and advises to switch to the latter.


marefr (approved):
If it works feel free to merge

PR #199, opened 2020-12-10
POC: Massive performance boost experiment with no full page reloads between renders

### LINKED ISSUE #124: Image renderer plugin: rendering horribly slow
**What would you like to be added**:

Improved image renderer plugin performance with panel rendering below 1s.

**Why is this needed**:

I'm running Grafana on an Intel-powered Synology NAS. Export of panels als PNGs or hot-linking to such PNGs is horribly slow as the time to generate them is between 5 and 15 seconds for a single panel. This is pretty much unusable except for batching purposes. 

While the box is not top-notch it has sufficient performance and 12GB of memory that suffices to run my other docker workload.

With rollout of the renderer plugin (https://github.com/grafana/grafana-image-renderer/issues/106) it would be great if the performance could be enhanced, e.g. by keeping a headless chrome running instead of starting per image.

torkelo:
Fixes #124

This is a very hacky POC to explore the biggest potential avenue for performance improvement, getting rid of full-page reloads between image renderings.

The first image takes the normal amount of time: 2-3 seconds 

After that, the image renderings take about 200 - 300ms (assuming the panel queries are very fast) 

This needs a lot more work to be production-ready. There could be big security issues with this approach that needs to be investigated. And more complex logic that validates that the rendered panel is for the same instance and org as the last render request, need a way to cache browser pages per org. 

Todo
* [ ] Address security concerns
* [ ] Cache pages per org & user
* [ ] Every 50x render (or some number) do a full page reload to avoid memory leaks causing issues
* [ ] Fix rendering different panel on same dashboard (should be easy) 
* [ ] Need not reuse for concurrent requests
* [ ] Needs updates to SoloPanelPage to support navigating between dashboards (needs cleanup), and panel on the same dashboard.

### DIFF
--- a/dev.json
+++ b/dev.json
@@ -19,10 +19,7 @@
   },
   "rendering": {
     "chromeBin": null,
-    "args": [
-      "--no-sandbox",
-      "--disable-setuid-sandbox"
-    ],
+    "args": ["--no-sandbox", "--disable-setuid-sandbox"],
     "ignoresHttpsErrors": false,
 
     "timezone": null,
@@ -34,7 +31,7 @@
     "maxHeight": 8000,
     "maxDeviceScaleFactor": 10,
 
-    "mode": "default",
+    "mode": "reusable",
     "clustering": {
       "mode": "browser",
       "maxConcurrency": 5
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "precommit": "npm run tslint & npm run typecheck",
     "watch": "tsc-watch --onSuccess \"node build/app.js server --config=dev.json\"",
     "build": "tsc",
-    "start": "node build/app.js --config=dev.json"
+    "start": "node build/app.js server --config=dev.json"
   },
   "dependencies": {
     "@grpc/grpc-js": "^1.0",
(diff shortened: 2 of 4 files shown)


xf-:
nice workaround :+1:

ryantxu:
The blocker with this approach is authentication managed with `renderKey` -- since it is a cookie (and must be a cookie) we can't share a page/context across multiple requests with different users.

In #314 @ArturWierzbicki has extended the concurrency mode so that requests with the same render key will keep the same session/cache etc.  But this will not yet help the general case.

ArturWierzbicki:
> The blocker with this approach is authentication managed with `renderKey` -- since it is a cookie (and must be a cookie) we can't share a page/context across multiple requests with different users.
> 
> In #314 @ArturWierzbicki has extended the concurrency mode so that requests with the same render key will keep the same session/cache etc. But this will not yet help the general case.

Created a working PoC in grafana-core for the general case, ie. when a single user issues multiple render requests in a short time window. 
The implementation is pretty simple - it leverages the new [contextPerRenderKey](https://github.com/grafana/grafana-image-renderer/pull/314) mode and persists the combination of `renderKey`/`userId`/`orgId`/`orgRole` in the DB. Keys are valid for 10 minutes unless refreshed via a new render request from the same user.

PR #214, opened 2021-04-23
Chore: bump Node to LTS (14.16.1)

theneva:
Hi! :wave:

After bumping our own stuff to Node 14, I noticed that this image is still on Node 12.

I think I've changed all the necessary things – relying on the implicit `latest` tag in the Dockerfiles to get at least Node 14.16.1 – and the app starts up fine.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
   },
   "devDependencies": {
     "@types/express": "^4.11.1",
-    "@types/node": "^10.0.9",
+    "@types/node": "^14.14.41",
     "husky": "^3.1.0",
     "lint-staged": "^9.5.0",
     "pkg": "4.4.8",
@@ -64,9 +64,9 @@
   },
   "bin": "build/app.js",
   "engines": {
-    "node": ">=12 <13"
+    "node": ">=14 <15"
   },
   "volta": {
-    "node": "12.19.0"
+    "node": "14.16.1"
   }
 }
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,7 +13,7 @@ aliases:
 defaults: &defaults
   working_directory: ~/grafana-image-renderer
   docker:
-    - image: circleci/node:12.13.1-stretch
+    - image: circleci/node:14.16.1-stretch
 
 commands:
   install-grabpl:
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -37,7 +37,7 @@
   },
   "devDependencies": {
     "@types/express": "^4.11.1",
-    "@types/node": "^10.0.9",
+    "@types/node": "^14.14.41",
```
theneva, line 40: 14.14.41 is the newest version, but I can't imagine there being any significant-to-this-project changes between 14.14 and 14.16 :smile:

package.json
```diff
@@ -64,9 +64,9 @@
   },
   "bin": "build/app.js",
   "engines": {
-    "node": ">=12 <13"
+    "node": ">=14 <15"
```
theneva, line 67: Node 15 reaches EOL on June 1st, so we might as well guard against it right away.

theneva:
I see that the build is failing, but I don't think I have permissions to see why 😞

aknuds1 (approved):
LGTM, WDYT @marefr?

marefr:
Have you tried building and running the executable plugins after this upgrade? Or only tested the docker image? I remember just going from Node 10 to Node 12 had a lot of issues with [pkg](https://www.npmjs.com/package/pkg)

package.json
theneva: Hm, maybe this is dumb since the code can also be used as a plugin, not just a Docker image. Should it be limited to any 12.x.x or 14.x.x?

theneva:
I only tested the Docker image since that's what I use – I didn't even consider that the other modes might break

theneva:
If it's not too hard to explain: How do I run the executable plugin(s) for testing?

aknuds1:
CI fails because of a bug in the CI config BTW, so it only works for non-forked PRs.

marefr:
Please try merging with master to resolve build errors related to plugin signing

AgnesToulet:
Thank you for your contribution! Your work has been integrated in this PR (https://github.com/grafana/grafana-image-renderer/pull/218) along with other dependencies update so I'm closing this PR.

PR #217, opened 2021-05-05
Add CSV export feature

AgnesToulet:
Extend the existing `render` function to be able to render panels as CSVs and not PNGs.
Plus, extend the `RenderResponse` to be able to return a file name, used to have access to the original download file name. 

**Open questions:**
- I'm not sure if the `yarn.lock` file contains the right information as I did a lot of tests and may have pushed some wrong content. How can I clean it up? 
- I'm not sure about extending the existing `render` function or make a new one. There is a lot in common between the CSV rendering and the PNG rendering.

### DIFF
--- a/src/browser/browser.ts
+++ b/src/browser/browser.ts
@@ -122,7 +122,8 @@ export class Browser {
       page = await browser.newPage();
       this.addPageListeners(page);
 
-      return await this.takeScreenshot(page, options);
+      return await this.exportCSV(page, options);
+      // return await this.takeScreenshot(page, options);
     } finally {
       if (page) {
         this.removePageListeners(page);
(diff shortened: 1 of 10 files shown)


Clarity-89:
> I'm not sure if the yarn.lock file contains the right information as I did a lot of tests and may have pushed some wrong content. How can I clean it up?

If you run `yarn install` it should cleanup the `yarn.lock` according to the package.json.

src/browser/browser.ts
```diff
@@ -122,7 +136,13 @@ export class Browser {
       page = await browser.newPage();
       this.addPageListeners(page);
 
-      return await this.takeScreenshot(page, options);
+      switch (options.renderType) {
+        case RenderType.CSV:
+          return await this.exportCSV(page, options);
```
Clarity-89, line 141:
```suggestion
          return this.exportCSV(page, options);
```
`await` is redundant if you're just returning the promise.

src/browser/browser.ts
```diff
@@ -122,7 +136,13 @@ export class Browser {
...
+        case RenderType.PNG:
+        default:
+          return await this.takeScreenshot(page, options);
```
Clarity-89, line 144:
```suggestion
          return this.takeScreenshot(page, options);
```

src/browser/browser.ts
```diff
@@ -134,6 +154,67 @@ export class Browser {
...
+    await page.setCookie({
+      name: 'renderKey',
+      value: options.renderKey,
+      domain: options.domain,
+    });
+
+    if (options.headers && Object.keys(options.headers).length > 0) {
+      this.log.debug(`Setting extra HTTP headers for page`, 'headers', options.headers);
+      await page.setExtraHTTPHeaders(options.headers);
+    }
```
Clarity-89, line 170: Wonder if we could abstract this duplicate logic to a function?

Clarity-89 (commented):
Code looks, good. Will test the behaviour in a bit.

joanlopez (approved):
AFAICT the code looks good to me!

PS: I'll try to set it up locally meanwhile to test the behavior

src/browser/browser.ts
AgnesToulet: Actually, I just discovered that this is not redundant in a try/(catch)/finally clause. If we remove this, the finally code will be executed before the function ends and the page will be closed too soon.

Clarity-89: True, good point. Didn't notice this was used inside `try/catch`.

Clarity-89 (approved):

marefr (approved):
LGTM

Issue #223, opened 2021-05-12
Panels are not using the selected timezone from the url query


Bujupah:
**Environment**:
- Grafana Image Renderer version: 2.1.1
- Grafana version: 7.3.1
- Installed plugin or remote renderer service: Remote
- OS Grafana Image Renderer is installed on: Windows/Linux
- User OS & Browser: Windows/Linux | Chromium/Chrome

**Issue**:
I tried rendering a panel with a given TZ as query param
http://localhost:3000/render/d-solo/DRoB8klMk/_?orgId=1&panelId=20&width=1000&height=500&tz=Africa/Tunis

Fully aware that the timezone format should be in this list [ICUs metaZones.txt](https://source.chromium.org/chromium/chromium/src/+/master:third_party/icu/source/data/misc/metaZones.txt)

but it seems like it was always falling back to host machine TZ and the puppeteer is not using the TimeZone from the env.TZ value...

**config.json file**:
```json
{
   ...
  "rendering": {
    "chromeBin": null,
    "args": [
      "--no-sandbox"
    ],
    "ignoresHttpsErrors": true,

    "timezone": null,
    "acceptLanguage": null,
    "width": 1000,
    "height": 500,
    "deviceScaleFactor": 1,
    "maxWidth": 3080,
    "maxHeight": 3000,
    "maxDeviceScaleFactor": 3,

    "mode": "default",
    "clustering": {
      "mode": "clustered",
      "maxConcurrency": 10
    },

    "verboseLogging": false,
    "dumpio": false
  }
}
```

Bujupah:
Will raise a PR for a fix

Bujupah:
PR #224

Bujupah:
@marefr @torkelo @aknuds1 some check this out please 🔢 
Want to know if I am using tz wrongly

AgnesToulet:
Closed by https://github.com/grafana/grafana-image-renderer/pull/224

PR #224, opened 2021-05-12
Fix to render panels and follow the selected timezone from URL query if specified.

Bujupah:
Related to issue #223 

Fixing the rendered panel timezone with a custom TZ from URL query.

Need someone to produce that and check if timezone is failing too, if so review this PR and check it too. this has fixed my issue.


Bujupah:
Followed this issue in puppeteer repo
https://github.com/puppeteer/puppeteer/issues/3325

AgnesToulet:
Thanks for this PR! I did some tests and ran into the same issue. This PR fixes it but I have a few comments:
- Setting the env variable in the LaunchOptions seems to not have any effect so it may be best to remove the code from the `getLauncherOptions` and move it into a separate function like this: 
```
  async setTimezone(page, options) {
    const timezone = options.timezone || this.config.timezone;
    if (timezone) {
      await page.emulateTimezone(timezone);
    }
  }
```
@marefr WDYT? Do you know if it used to work?
- It also needs to be added in the new `renderCSV` function to be complete.

Bujupah:
yes! It kinda needs some refactoring tbh

marefr:
> Setting the env variable in the LaunchOptions seems to not have any effect so it may be best to remove the code from the getLauncherOptions and move it into a separate function like this:

I'm pretty sure it did work back in the days. But if this works all is good 👍

Bujupah:
> Thanks for this PR! I did some tests and ran into the same issue. This PR fixes it but I have a few comments:
> 
> * Setting the env variable in the LaunchOptions seems to not have any effect so it may be best to remove the code from the `getLauncherOptions` and move it into a separate function like this:
> 
> ```
>   async setTimezone(page, options) {
>     const timezone = options.timezone || this.config.timezone;
>     if (timezone) {
>       await page.emulateTimezone(timezone);
>     }
>   }
> ```
> 
> @marefr WDYT? Do you know if it used to work?
> 
> * It also needs to be added in the new `renderCSV` function to be complete.

I updated my PR with this function

AgnesToulet:
Sorry for the delay, thanks for the update! Could you also add it to the [`renderCSV`](https://github.com/grafana/grafana-image-renderer/blob/master/src/browser/browser.ts#L233) function please to have the same behavior? 

And also, could you merge with master to fix the issues with the pipeline? Thanks!

Bujupah:
Thanks @AgnesToulet
I made that change

AgnesToulet (approved):

PR #232, opened 2021-05-19
CSV: replace rename file with copy and unlink

AgnesToulet:
Rename can’t be used to rename across devices so this PR replaces `renameSync` with `copyFileSync` and `unlinkSync`.


xlson (approved):
I don't know much about nodejs but from my basic understanding that looks good :)

Clarity-89 (approved):

Issue #228, opened 2021-05-18
Remove support for deprecated backend plugin protocol version


marefr:
**What would you like to be added**:

Remove support for deprecated backend plugin protocol version. At same time limit the supported Grafana version to greater or equal than v7.0, see grafanaDependency https://grafana.com/docs/grafana/latest/developers/plugins/metadata/#dependencies

**Why is this needed**:

Removed support in Grafana via  https://github.com/grafana/grafana/issues/33959.

Basically needs to remove https://github.com/grafana/grafana-image-renderer/blob/master/src/plugin/v1/grpc_plugin.ts and any code using this.

PR #233, opened 2021-05-19
Remove references to plugin V1 protocol

AgnesToulet:
Remove references to plugin V1 protocol as it's not supported anymore by Grafana: see https://github.com/grafana/grafana/issues/33959

Closes https://github.com/grafana/grafana-image-renderer/issues/228

**Question**: I'm not sure if the `coreProtocolVersion` should be updated as well? It's only displayed when the plugin starts: https://github.com/grafana/grafana-image-renderer/search?q=coreprotocolversion


plugin.json
```diff
@@ -24,6 +24,6 @@
     "updated": "2021-05-19"
   },
   "dependencies": {
-    "grafanaVersion": "5.x.x 6.x.x 7.x.x 8.x.x"
+    "grafanaVersion": "7.x.x 8.x.x"
```
marefr, line 27: Should use grafanaDependency with a proper semver string

src/app.ts
```diff
@@ -78,28 +74,7 @@ main().catch(err => {
...
-  }
-
-  if (env['GF_RENDERER_PLUGIN_CHROME_BIN']) {
-    config.rendering.chromeBin = env['GF_RENDERER_PLUGIN_CHROME_BIN'];
-  }
-
-  if (env['GF_RENDERER_PLUGIN_VERBOSE_LOGGING']) {
-    config.rendering.verboseLogging = env['GF_RENDERER_PLUGIN_VERBOSE_LOGGING'] === 'true';
-  }
+  config.rendering.timezone = env['TZ'];
```
marefr, line 77: Any reason keeping this one? Or maybe move it as fallback to https://github.com/grafana/grafana-image-renderer/blob/1d6e373363373ee4a4390979d031bc9243ce69e9/src/plugin/v2/grpc_plugin.ts#L204 ?

marefr (changes requested):
Some minor things left. Otherwise looks great 👍

marefr (approved):
LGTM

PR #246, opened 2021-06-08
Upgrade dependencies

Clarity-89:
- Upgrade puppeteer, typescript and prettier to the latest versions

Tested on Mac and it builds and works ok. Could use more tests on other OS probably. 

Also did some brief profiling and it seems like total rendering time per panel is decreased by ~400ms, could be due to the newer versions of Chronium used.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -18,8 +18,8 @@ async function main() {
     const logger = new PluginLogger();
     const config: PluginConfig = defaultPluginConfig;
     populatePluginConfigFromEnv(config, env);
-
     if (!config.rendering.chromeBin && (process as any).pkg) {
+      //@ts-ignore
       const parts = puppeteer.executablePath().split(path.sep);
       while (!parts[0].startsWith('chrome-')) {
         parts.shift();
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "pkg": "^5.1.0",
     "prettier": "2.2.1",
     "tsc-watch": "^4.2.3",
-    "typescript": "4.2.4"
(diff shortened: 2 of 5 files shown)


src/app.ts
```diff
@@ -18,8 +18,8 @@ async function main() {
     const logger = new PluginLogger();
     const config: PluginConfig = defaultPluginConfig;
     populatePluginConfigFromEnv(config, env);
-
     if (!config.rendering.chromeBin && (process as any).pkg) {
+      //@ts-ignore
```
Clarity-89, line 22: This seems like seem issue with types, `executablePath` works but TS complains that it doesn't exist.

AgnesToulet:
Did some tests on Windows and it crashes when running in plugin mode because of this issue: https://github.com/puppeteer/puppeteer/issues/6563.
The fix is to close all the pages before closing the browser in the `getBrowserVersion` function:
```
  async getBrowserVersion(): Promise<string> {
    let browser;

    try {
      const launcherOptions = this.getLauncherOptions({});
      browser = await puppeteer.launch(launcherOptions);
      return await browser.version();
    } finally {
      if (browser) {
        const pages = await browser.pages();
        await Promise.all(pages.map(page =>page.close()));
        await browser.close();
      }
    }
  }
```

I need to do more tests with the different options of the plugin but it looks good otherwise!

Clarity-89:
@AgnesToulet are you able to fix the Windows issue? It's a bit hard for me to do as I don't have Windows machine setup.

DanCech (approved):
LGTM

PR #249, opened 2021-06-08
Chore: Don't fail PR build if plugin signing fails fix

AgnesToulet:
Follow-up of https://github.com/grafana/grafana-image-renderer/pull/216

The default bash used in CircleCi is `/bin/bash -eo pipefail` and `-e` means:
> Exit immediately if a pipeline (which may consist of a single simple command), a subshell command enclosed in parentheses, or one of the commands executed as part of a command list enclosed by braces exits with a non-zero status.

As explained here: https://circleci.com/docs/2.0/configuration-reference/#default-shell-options

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -88,6 +88,7 @@ jobs:
       - run:
           name: Create and sign plugin manifest
           command: |
+            set +e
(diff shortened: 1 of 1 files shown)


marefr (approved):
LGTM

PR #257, opened 2021-06-29
renderer using root user instead of non-root

VishalLahane:
- Current user
```
/usr/src/app # id
uid=0(root) gid=0(root) groups=0(root),1(bin),2(daemon),3(sys),4(adm),6(disk),10(wheel),11(floppy),20(dialout),26(tape),27(video)

```

After change 
```
 id
uid=1000(node) gid=1000(node) groups=1000(node),1000(node)

```


VishalLahane:
As there already PR for same

Issue #229, opened 2021-05-19
Browser page crashed


jgournet:
**What happened**:
1) overall setup is working well: image renderer works and sends nice images to slack.
2) Then, "something" happens, and from there on, image-renderer keeps on crashing. A pod restart solves the issue for any time between a few hours and 2 weeks.
Logs, including 1 successfull call  at the top (just to show that it does work fine)
```
 2021-05-18T22:26:33.755623434Z {"message":"::ffff:127.0.0.1 - - [18/May/2021:22:26:33 +0000] \"GET /render?deviceScaleFactor=1.000000&domain=localhost&encoding=&height=500&renderKey=QAWfFera3MASlHo2K9CsSfU76HXIMt74&timeout=15&timezone=&url=http%3A%2F%2Flocalhost%3A3000%2Fd-solo%2F63dae39ae810539b3108233c3749bf114998da82%2Fsbr2-compliance-production%3ForgId%3D1%26panelId%3D29%26render%3D1&width=1000 HTTP/1.1\" 200 33110 \"-\" \"Grafana/7.1.5\"\n","level":"debug"}
 2021-05-18T22:26:33.755729023Z {"url":"http://localhost:3000/d-solo/63dae39ae810539b3108233c3749bf114998da82/sbr2-compliance-production?orgId=1&panelId=29&render=1","level":"debug","message":"Connection closed"}
 2021-05-18T22:31:27.805524963Z {"url":"http://localhost:3000/d-solo/63dae39ae810539b3108233c3749bf114998da82/sbr2-compliance-production?orgId=1&panelId=29&render=1","level":"debug","message":"Render request received"}
 2021-05-18T22:31:31.096108470Z {"file":"/tmp/9ef11c5d.png","level":"debug","message":"Deleting temporary file"}
[...]

PR #262, opened 2021-07-30
Add --disable-gpu in the default Chromium args

AgnesToulet:
While doing some load testing of the application with the `default` rendering mode, I noticed high CPU and memory slowy going to the roof after the tests were finished. I narrowed down the issue to the `newPage` function that was (for some requests) hanging forever, preventing the browser to be closed and causing these issues. 

It seems like adding the `--disable-gpu` flag to the Chromium args fixed the issue.

Should fix https://github.com/grafana/grafana-image-renderer/issues/229

### DIFF
--- a/default.json
+++ b/default.json
@@ -19,9 +19,7 @@
   },
   "rendering": {
     "chromeBin": null,
-    "args": [
-      "--no-sandbox"
-    ],
+    "args": ["--no-sandbox", "--disable-gpu"],
     "ignoresHttpsErrors": false,
 
     "timezone": null,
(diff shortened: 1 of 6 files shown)


marefr (approved):
LGTM I guess, but wonder what possible issues this might cause for others basically switching from hardware acceleration to software rendering if I understand it correctly? Guess just try, but make sure to make the change clear in the change log should anyone have issues with it.

AgnesToulet:
It's updated only with a new release. Next release should happen this week.

PR #263, opened 2021-07-30
Add browser timing metrics

AgnesToulet:
Add back browser timing metrics to understand what actions are taking the most time, see evolution over time and help troubleshoot performance issues.


Clarity-89 (approved):
Looks good!

PR #273, opened 2021-08-25
Docker: revert changes on the working directory

AgnesToulet:
Use the same working directory as before to not break existing Docker set-ups.

I'm no Docker expert so if there is a benefit to use `/usr/share/grafana` instead of `/usr/src/app`, we can keep it but if there isn't, then, it's best to not introduce a breaking change for Docker users.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,8 +32,11 @@ ARG GF_UID="472"
 ARG GF_GID="472"
 ENV GF_PATHS_HOME="/usr/src/app"
 
+ WORKDIR $GF_PATHS_HOME
+
 RUN addgroup -S -g $GF_GID grafana && \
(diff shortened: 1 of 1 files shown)


joanlopez (approved):
It looks good AFAICT, but let's wait for an additional review 🙏🏻

xlson (commented):
I would suggest keeping the changes as they were for the most part but to change line 33 to
```
ENV GF_PATHS_HOME="/usr/src/app"
```

This reverts the workdir configuration but keeps the rest of the changes intact, making it possible to select a custom working directory if you want that for some reason.

sakjur (approved):

xlson (approved):

DanCech (approved):

Issue #234, opened 2021-05-20
Instrumentation: Add gauge of total number of requests in flight. 


bergquist:
Exposing a gauge with the total number of requests currently being served is a great way to identify bottlenecks. 

Ex https://github.com/grafana/grafana/blob/main/pkg/middleware/request_metrics.go#L26

PR #281, opened 2021-09-07
Instrumentation: Add gauge of total number of requests in flight

AgnesToulet:
Add gauge of total number of requests in flight (only for `/render/*` routes). 

Closes https://github.com/grafana/grafana-image-renderer/issues/234

### DIFF
--- a/src/service/metrics_middleware.ts
+++ b/src/service/metrics_middleware.ts
@@ -1,14 +1,12 @@
 import * as promBundle from 'express-prom-bundle';
+import * as promClient from 'prom-client';
+import * as onFinished from 'on-finished';
+import express = require('express');
+
 import { MetricsConfig } from '../config';
 import { Logger } from '../logger';
 
-export const metricsMiddleware = (config: MetricsConfig, log: Logger) => {
-  if (!config.enabled) {
-    return (req, res, next) => {
-      next();
-    };
-  }
-
+export const setupHttpServerMetrics = (app: express.Express, config: MetricsConfig, log: Logger) => {
   log.info(
     'Metrics enabled',
     'collectDefaultMetrics',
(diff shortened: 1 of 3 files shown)


src/service/metrics_middleware.ts
```diff
@@ -17,13 +15,15 @@ export const metricsMiddleware = (config: MetricsConfig, log: Logger) => {
     config.requestDurationBuckets.join(',')
   );
 
+  const excludeRegExp = /^((?!(render)).)*$/;
```
bergquist, line 18: how come these routes are excluded?

src/service/metrics_middleware.ts
```diff
@@ -39,6 +39,29 @@ export const metricsMiddleware = (config: MetricsConfig, log: Logger) => {
...
+  return (req, res, next) => {
+    const path = req.originalUrl || req.url;
+    if (path.match(excludeRegExp)) {
+      return next();
+    }
+
+    httpRequestsInFlight.inc();
+    onFinished(res, () => {
+      httpRequestsInFlight.dec();
+      next();
```
bergquist, line 62: should `next()` really be called here? I suspect that the `next()` call on 65 would be enough to progress. But it was a long time since I wrote nodejs code :)

src/service/metrics_middleware.ts
AgnesToulet: It's to exclude all routes that don't contain `render`. I'm not sure it's really useful currently but it could be in the future (right now, the only route that is excluded is `GET /`)

src/service/metrics_middleware.ts
AgnesToulet: Yes you're right, it's a mistake, I'll remove it.

Clarity-89:
Should I be able to see these metrics somewhere or are they just logged into console? I have this in my dev.json (default setup):
``` 
"metrics": {
  "enabled": true,
  "collectDefaultMetrics": true,
  "requestDurationBuckets": [1, 5, 7, 9, 11, 13, 15, 20, 30]
},
```

AgnesToulet:
Yes you can see them by calling the `/metrics` endpoint of the image renderer.

Clarity-89 (approved):
Looks good!

PR #293, opened 2021-10-04
Chore: Fix eslint usage

AgnesToulet:
- Move `eslint` into dev dependencies
- Bump eslint from 7.13.0 to 7.32.0
- Extends `@grafana/eslint-config` instead of having plugin rules
- Fix `eslint` usage


package.json
```diff
@@ -37,8 +36,17 @@
...
     "@types/express": "^4.11.1",
     "@types/node": "^14.14.41",
+    "@typescript-eslint/eslint-plugin": "^4.32.0",
+    "@typescript-eslint/parser": "^4.32.0",
+    "eslint": "^7.32.0",
+    "eslint-config-prettier": "^8.3.0",
+    "eslint-plugin-jsdoc": "^36.1.0",
+    "eslint-plugin-prettier": "^4.0.0",
+    "eslint-plugin-react": "^7.26.1",
+    "eslint-plugin-react-hooks": "^4.2.0",
```
Clarity-89, line 49: Not sure we need these since we're not using react here, the same goes for jsdoc plugin.

Clarity-89 (approved):
Looks good, just a minor comment.

package.json
AgnesToulet: I agree, we don't really need them, but they are required when using this `"extends": ["@grafana/eslint-config"]`.

I honestly wondered if it was worth it or if we should use our own config and preferred to use the Grafana config and keep these extra dependencies. WDYT?

Clarity-89: Ah I see, I guess we can keep these extra dependencies, they're dev dependencies anyways, so we don't need to maintain a separate config for the image renderer.

PR #299, opened 2021-10-26
Fix check condition to avoid timeouts in invalid panels

spinillos:
When we are trying to render a panel that isn't able to retrieve the information from the datasource, `panelsRendered` never exists.

It means, that we are waiting to fit a condition that is never met. For this reason, image renderer waiting until the timeout is reached.

Checking if `panelsRendered` exists, takes the screenshot of the invalid panel avoiding the timeout.


AgnesToulet (approved):
LGTM!

Issue #279, opened 2021-09-01
Configuration: Add a setting to override puppeteer-cluster timeout 


AgnesToulet:
**What would you like to be added**:
Add a new configuration field to override the default timeout for the `clustered` mode.

**Why is this needed**:
Each rendering request can send a [timeout option](https://github.com/grafana/grafana-image-renderer/blob/master/src/browser/browser.ts#L25) that will be used when navigating to a page, waiting for panels to be rendered or taking a sreenshot.
However, in `clustered` mode, a timeout of 30 seconds is already set when launching the cluster (see https://github.com/thomasdondorf/puppeteer-cluster#clusterlaunchoptions). So, even if the request timeout is over 30 seconds, the rendering request will be stopped after 30 seconds by the underlying cluster library.

Bujupah:
I am facing an issue with timeout related to puppeteer-cluster 

```
Error: Timeout hit: 30000
    at Object.<anonymous>(/usr/src/app/node_modules/puppeteer-cluster/dist/util.js:69:23)
    at Generator.next (<anonymous>)
    at fulfilled (/usr/src/app/node_modules/puppeteer-cluster/dist/util.js:5:58)
    at runNextTicks (internal/process/task_queues.js:60:5)
    at listOnTimeout (internal/timers.js:526:9)
    at processTimers (internal/timers.js:500:7)"
```

I tried to increase it to 60s but same issue is appearing... 
Any idea what could be the reason?
is there any suggested value for that timeout?

PR #303, opened 2021-11-10
Configuration: add timeout setting for clustered mode

AgnesToulet:
Add timeout setting that overrides puppeteer-cluster timeout. 

Fixes https://github.com/grafana/grafana-image-renderer/issues/279

### DIFF
--- a/default.json
+++ b/default.json
@@ -34,7 +34,8 @@
     "mode": "default",
     "clustering": {
       "mode": "browser",
-      "maxConcurrency": 5
+      "maxConcurrency": 5,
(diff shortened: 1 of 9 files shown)


spinillos (approved):
Looks like fine

xlson (approved):
Works for me :+1: 

2021-11-17-112606_770x422_scrot

One thing that doesn't make sense to me is the error we return. Returning a 500 from the image renderer for a timeout is pretty dirty. It would make more sense to me to return an image letting the user know they hit a timeout.

PR #308, opened 2021-12-01
Types: improve options types

ryantxu:
There are currently two types `RenderOptions` and `RenderCSVOptions` however, CSV is just a subset of the other types.  In addition, the HTTP server has an identical type but it is called `RenderRequest`  -- grpc plugin has a similar type -- but headers is string[] rather than string so :man_shrugging: 

This PR:
1. adds a types.ts file to root and puts RenderOptions and ImageRenderOptions there
2. avoids using `any` in function calls
3. use the same options type for browser and server

This has no behavior changes


AgnesToulet (approved):
LGTM, thanks for the improvement!

PR #312, opened 2022-01-06
Previews: support creating thumbnails and full-height screenshots

ryantxu:
This PR updates the image renderer to support creating thumbnails from a requested page.  Rather than adding a new parameter, this is reusing the scale/deviceScaleFactor parameter and using negative numbers.

For the request:
http://localhost:3000/render/d/zNigXIonz/heatmap?kiosk&orgId=1&width=320&height=240&scale=-5

It will render an image a 320*5 and then scale it back down to 320px:

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 3.4.0 (--)
+- Support full height dashboards and scaled thumbnails [#312](https://github.com/grafana/grafana-image-renderer/pull/312)
+
 ## 3.3.0 (2021-11-18)
 
 - Chore: Bump pkg from 5.3.3 to 5.4.1 [#305](https://github.com/grafana/grafana-image-renderer/pull/305), [AgnesToulet](https://github.com/AgnesToulet)
--- a/scripts/pkg.js
+++ b/scripts/pkg.js
@@ -23,5 +23,9 @@ const archTransform = {
 platform = platformTransform[platform] || platform;
 arch = archTransform[arch] || arch;
 const outputPath = "dist/" + (process.argv[3] || `plugin-${archArg}`);
+const outputNodeModules = `${outputPath}/node_modules`
 
 childProcess.execSync(`"./node_modules/.bin/pkg" -t node14-${platform}-${arch} . --out-path ${outputPath}`, {stdio: 'inherit'});
+
+childProcess.execSync(`rm -rf ${outputNodeModules} && mkdir -p ${outputNodeModules}/sharp`)
+childProcess.execSync(`cp -RP ./node_modules/sharp/build ${outputNodeModules}/sharp && cp -RP ./node_modules/sharp/vendor ${outputNodeModules}/sharp`)
--- a/scripts/pkg.js
+++ b/scripts/pkg.js
@@ -1,4 +1,5 @@
 const childProcess = require('child_process');
+const fs = require('fs')
 
 const archArg = process.argv[2];
 let [
@@ -27,5 +28,6 @@ const outputNodeModules = `${outputPath}/node_modules`
 
 childProcess.execSync(`"./node_modules/.bin/pkg" -t node14-${platform}-${arch} . --out-path ${outputPath}`, {stdio: 'inherit'});
 
-childProcess.execSync(`rm -rf ${outputNodeModules} && mkdir -p ${outputNodeModules}/sharp`)
+childProcess.execSync(`rm -rf ${outputNodeModules}`)
+fs.mkdirSync(`${outputNodeModules}/sharp`, {recursive: true})
 childProcess.execSync(`cp -RP ./node_modules/sharp/build ${outputNodeModules}/sharp && cp -RP ./node_modules/sharp/vendor ${outputNodeModules}/sharp`)
--- a/scripts/pkg.js
+++ b/scripts/pkg.js
@@ -23,10 +23,15 @@ const archTransform = {
 
 platform = platformTransform[platform] || platform;
 arch = archTransform[arch] || arch;
+
+if(platform === 'macos' && (arch.includes('arm'))) {
+  arch = 'arm64'
+}
+
 const outputPath = "dist/" + (process.argv[3] || `plugin-${archArg}`);
 const outputNodeModules = `${outputPath}/node_modules`
 
-childProcess.execSync(`"./node_modules/.bin/pkg" -t node14-${platform}-${arch} . --out-path ${outputPath}`, {stdio: 'inherit'});
+childProcess.execSync(`"./node_modules/.bin/pkg" -t node14-${platform}-${arch} . --out-path ${outputPath} --no-native-build`, {stdio: 'inherit'});
 
 childProcess.execSync(`rm -rf ${outputNodeModules}`)
 fs.mkdirSync(`${outputNodeModules}/sharp`, {recursive: true})
--- a/src/browser/browser.ts
+++ b/src/browser/browser.ts
@@ -351,10 +351,12 @@ export class Browser {
 
     const screenshotFn = () =>
       page.screenshot({ path: options.filePath, fullPage: options.fullPageImage, captureBeyondViewport: options.fullPageImage });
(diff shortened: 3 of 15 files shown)


src/browser/browser.ts
```diff
@@ -230,10 +248,27 @@ export class Browser {
       this.log.debug('Taking screenshot', 'filePath', options.filePath);
     }
 
+    if (options.fullPageImage) {
+      this.log.debug('TODO: take full screen image', 'url', options.url);
```
ryantxu, line 252: @ArturWierzbicki -- how involved was this process?  If easy to port, can we try that?

src/browser/browser.ts
```diff
@@ -230,9 +349,25 @@ export class Browser {
...
-    }, 'screenshot');
+    const screenshotFn = () =>
+      page.screenshot({ path: options.filePath, fullPage: options.fullPageImage, captureBeyondViewport: options.fullPageImage });
+    await this.withTimingMetrics(
+      scrollResult.scrolled ? this.withFullPageViewport(screenshotFn, page, options, scrollResult.scrollHeight) : screenshotFn,
+      'screenshot'
+    );
+
+    if (options.scaleImage) {
+      const scaled = uniqueFilename(os.tmpdir()) + '.webp';
```
ryantxu, line 360: @ArturWierzbicki -- I think we should change this back to .png for now.  Perhaps we can introduce a new 'format' parameter to change it to webp

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 3.4.0 (--)
+- Support full height dashboards and scaled thumbnails [#312](https://github.com/grafana/grafana-image-renderer/pull/312)
```
ryantxu, line 2: @AgnesToulet what is the process for documenting the changes here?

The behavior changes are:
1. negative scale will create a scaled down thumbnail
2. height=-1 will create an image as tall as it needs to be based on the webpage size

src/config.ts
```diff
@@ -23,6 +23,7 @@ export interface RenderingConfig {
   verboseLogging: boolean;
   dumpio: boolean;
   timingMetrics: boolean;
+  headed?: boolean;
```
AgnesToulet, line 26: I think it would be nice to add this in the `default.json` and `dev.json` files set to false so users and developers know that it exists.

AgnesToulet (approved):
LGTM! Left a couple of questions and suggestions but all very subjective and nothing major. 

Also, I wonder if it would be worth to support the full page option without the dashboard being in kiosk mode (right now, dashboards can be a bit cut at the end without the kiosk mode).

src/types.ts
```diff
@@ -19,4 +19,9 @@ export interface ImageRenderOptions extends RenderOptions {
   width: string | number;
   height: string | number;
   deviceScaleFactor?: string | number;
+
+  // Runtime options derived from the input
+  fullPageImage?: boolean;
+  scrollDelay?: number;
+  scaleImage?: number;
```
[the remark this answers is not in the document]
ArturWierzbicki: I assume that's an easy way to make it work without changing the API version? @ryantxu

src/browser/browser.ts
```diff
@@ -230,9 +349,27 @@ export class Browser {
...
 
-    await this.withTimingMetrics<void>(() => {
-      return page.screenshot({ path: options.filePath });
-    }, 'screenshot');
+    const screenshotFn = () =>
+      page.screenshot({ path: options.filePath, fullPage: options.fullPageImage, captureBeyondViewport: options.fullPageImage });
+    await this.withTimingMetrics(
+      scrollResult.scrolled ? this.withFullPageViewport(screenshotFn, page, options, scrollResult.scrollHeight) : screenshotFn,
+      'screenshot'
+    );
```
AgnesToulet, line 357: *This comment got lost when submitting the review...*

This is a bit hard to read and I don't think we need to set back the viewport to its original state in case of a full page screenshot. Could we simplify with something like this?
```suggestion
    await this.withTimingMetrics(async () => {
      if (scrollResult.scrolled) {
        await this.setViewport(page, {
          ...options,
          height: scrollResult.scrollHeight,
        });
      }

      return page.screenshot({ path: options.filePath, fullPage: options.fullPageImage, captureBeyondViewport: options.fullPageImage });
    }, 'screenshot');
```

ArturWierzbicki: Removed. The viewport reset was there because I ported the code from the hackathon version, where we were also reusing the page to take screenshots of each panel

src/types.ts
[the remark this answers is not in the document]
ryantxu: I think reusing `deviceScaleFactor` is ok because:
1. the external param for this is already called "scale"! :)
2. would never want deviceScaleFactor != 1 and scale !=1 at the same time

In general, I don't really understand  why `deviceScaleFactor` exists -- i *think* you get identical results just scaling the image width/height.

ArturWierzbicki:
@AgnesToulet fixed non-kiosk mode in https://github.com/grafana/grafana-image-renderer/pull/312/commits/efa7e73b2ca1c4c5099e5bc2802df53da79ffd27

ArturWierzbicki:
There is still one issue reported by `prebuild-install`: 

image

```
prebuild-install WARN This package does not support N-API version v14.18.3 
prebuild-install WARN install No prebuilt binaries found (target=v14.18.3 runtime=napi arch=x64 libc= platform=linux)
```

Which is weird, because Sharp requires NAPI version >= 5 which is supported by node 14/16:  https://nodejs.org/api/n-api.html#n_api_node_api_version_matrix

CHANGELOG.md
achatterjee-grafana: Sounds good @AgnesToulet. Let me know if you need any help.

ArturWierzbicki:
The only way I found to make it work is to manually copy sharp binary/`libvips` lib & downgrade `pkg` to 4.4.9 - which is right before native support for `.node` files (https://github.com/vercel/pkg/releases/tag/4.5.0). Works on linux, will test on MacOS as well

ArturWierzbicki:
Packaging on M1 MacOS doesn't work, neither on this branch (because 2.6.x `pkg-fetch` doesn't have precompiled arm nodejs) nor on `main`

ArturWierzbicki:
Tested on Windows x64 with CI-built binary https://app.circleci.com/pipelines/github/grafana/grafana-image-renderer/464/workflows/8d39e08f-a376-4d41-8c23-62b4f80c5562/jobs/1864/artifacts - works

AgnesToulet (approved):
LGTM! Thanks for the fix for sharp library. Left a few very minor suggestions and one comment to make the scripts work on Windows (the priority is to make it work on Linux but if, with some tweaks, we can make it work on both, it would be great!)

scripts/pkg.js
```diff
@@ -22,6 +22,15 @@ const archTransform = {
...
+  arch = 'arm64'
+}
+
 const outputPath = "dist/" + (process.argv[3] || `plugin-${archArg}`);
+const outputNodeModules = `${outputPath}/node_modules`
+
+childProcess.execSync(`"./node_modules/.bin/pkg" -t node14-${platform}-${arch} . --out-path ${outputPath} --no-native-build`, {stdio: 'inherit'});
 
-childProcess.execSync(`"./node_modules/.bin/pkg" -t node14-${platform}-${arch} . --out-path ${outputPath}`, {stdio: 'inherit'});
+childProcess.execSync(`rm -rf ${outputNodeModules} && mkdir -p ${outputNodeModules}/sharp`)
```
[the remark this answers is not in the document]
ArturWierzbicki: replaced `mkdir -p` with `fs.mkdirSync` and retested on linux & m1!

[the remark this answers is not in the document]
ArturWierzbicki: @AgnesToulet awesome! good to merge?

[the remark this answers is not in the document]
AgnesToulet: Yes, let's merge! 🎉

Issue #323, opened 2022-02-19
v3.4.0 error="*rendering.RenderingService run error: Unrecognized remote plugin message:..."


robonuka:
**What happened**: grafana-image-renderer v3.4.0 brings grafana down with error:

` EROR[02-19|12:33:23] Stopped background service               logger=server service=*rendering.RenderingService reason="Unrecognized remote plugin message: \n\nThis usually means that the plugin is either invalid or simply\nneeds to be recompiled to support the latest protocol."
INFO[02-19|12:33:23] Closing tracing                          logger=tracing
WARN[02-19|12:33:23] plugin failed to exit gracefully         logger=plugin.grafana-image-renderer
EROR[02-19|12:33:23] context canceled                         logger=ngalert
EROR[02-19|12:33:23] Server shutdown                          logger=server error="*rendering.RenderingService run error: Unrecognized remote plugin message: \n\nThis usually means that the plugin is either invalid or simply\nneeds to be recompiled to support the latest protocol."
*rendering.RenderingService run error: Unrecognized remote plugin message: 
This usually means that the plugin is either invalid or simply
needs to be recompiled to support the latest protocol. `

**What you expected to happen**: normal work as v3.3.0 does

**How to reproduce it (as minimally and precisely as possible)**: Just install v3.4.0 grafana-image-renderer plugin

**Anything else we need to know?**:

**Environment**: self-made docker image
- Grafana Image Renderer version: v3.4.0
- Grafana version: grafana/grafana:main or grafana/grafana:8.4.1
- Installed plugin or remote renderer service: plugin installed with
`grafana-cli --pluginsDir "$GF_PATHS_PLUGINS" --pluginUrl https://github.com/grafana/grafana-image-renderer/releases/latest/download/plugin-linux-x64-glibc-no-chromium.zip glibc-no-chromium.zip plugins install grafana-image-renderer`
- OS Grafana Image Renderer is installed on: Alpine (original OS in the grafana/grafana:XXX docker images)
- User OS & Browser: any (grafana container goes down while starting)
- Others: if I downgrade grafana-image-renderer plugin to v3.3.0 - everything is fine. Grafana works as usually.

ArturWierzbicki:
Hey @robonuka, thank you for reporting this issue! We have added dependency on `sharp` image processing library in 3.4.0 and that's the root cause of this issue. 

[...]

PR #314, opened 2022-01-10
Support new concurrency mode: `contextPerRenderKey` 

ArturWierzbicki:
PR adds support for a new concurrency mode, `contextPerRenderKey` which improves the performance of the dashboard crawl process without negatively affecting the existing use cases.

With `contextPerRenderKey` mode enabled, image-renderer will reuse the same [browser context](https://github.com/puppeteer/puppeteer/blob/main/docs/api.md#class-browsercontext) (created via CDP - https://chromedevtools.github.io/devtools-protocol/tot/Target/#method-createBrowserContext) for all requests with the same `domain` and `renderKey` within a certain timeframe. 

The dashboard crawler reuses the same `renderKey` for all renders within a single crawl thanks to the [concept of RenderingSession](https://github.com/grafana/grafana/pull/44098/files#diff-f412c5dc4b56f8c250a37c17e8291f74bd0ef8694afb1f8f4ce7d77368b414a1R67). All other use cases create a unique `renderKey` for each request, causing `contextPerRenderKey` mode to behave exactly the same as the existing `context` mode

The main benefits are:
- lower setup overhead for each render: new `browserContext` vs new page in existing `browserContext`
   - impact: ~~negligible
- lower CPU/mem overhead: `browserContext` is a separate browser window
   - impact: ~~negligible
- shared JS cache
   - impact: big! but depends on the network conditions

Shared JS cache is by far the biggest deal. Tests done in local env with 94 dashboards and 6 crawler threads - results are the average of 3 runs (low variance)

| network                     | context | contextPerRenderKey |
|-----------------------------|---------|---------------------|
| not throttled               | 60s     | 60s                 |
| throttled: max 4mbps, 100ms | 95s     | 70s                 |
| throttled: max 1mbps, 100ms | 190s    | 78s                 |

### to test:

1. check out latest `main` grafana branch
2. enable `dashboardPreviews` feature flag
3. configure grafana to use http version of image-renderer via `custom.ini` config
```
[rendering]
server_url = http://localhost:8081/render
```
4. go to _server admin -> stats and license_ and click `start`
   - image

PR #325, opened 2022-02-23
Replace sharp with jimp

ArturWierzbicki:
Fixes https://github.com/grafana/grafana-image-renderer/issues/323

`sharp` is [faster](https://sharp.pixelplumbing.com/performance#results), but its native dependencies & our plugin mechanism don't work well together - at least for the time being. The numbers in sharp's benchmark seem scary, but it looks like it's not going to be an issue in our case - the numbers I posted in https://github.com/grafana/grafana-image-renderer/pull/314 remain unchanged.


AgnesToulet (approved):
LGTM! I would delete all sharp-related things in the scripts to not make it more complicated than it is. It will always be available through Git if we want to use sharp again.

PR #326, opened 2022-02-23
Add darwin arm64

ArturWierzbicki:
targets https://github.com/grafana/grafana-image-renderer/pull/325

### DIFF
--- a/scripts/download_chromium.js
+++ b/scripts/download_chromium.js
@@ -37,7 +37,13 @@ browserFetcher
 
     let execPath = parts.join(path.sep);
 
-    child_process.execSync(`cp -RP ${execPath} ${outputPath}`);
+    if (platform === 'mac' && arch === 'arm64') {
+        // follow symlinks, dereference symlinks and copy them as files
+        child_process.execSync(`cp -LR ${execPath} ${outputPath}`);
+    } else {
+        // follow symlinks, copy them as symlinks
+        child_process.execSync(`cp -RP ${execPath} ${outputPath}`);
+    }
 
     console.log(`Chromium moved from ${execPath} to ${outputPath}/`);
     process.exit(0);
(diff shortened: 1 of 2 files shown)


scripts/download_chromium.js
```diff
@@ -37,7 +38,21 @@ browserFetcher
 
     let execPath = parts.join(path.sep);
 
-    child_process.execSync(`cp -RP ${execPath} ${outputPath}`);
+    if (platform === 'mac' && arch === 'arm64') {
+        // follow symlinks, dereference symlinks and copy them as files
+        child_process.execSync(`cp -LR ${execPath} ${outputPath}`);
+
+        const dsStore = '.DS_Store'
```
ArturWierzbicki, line 45: I think the problem with `.DS_Store` files (https://en.wikipedia.org/wiki/.DS_Store) is the last one we need to solve before we can add the `darwin-arm64` target. 

Lets say we are packaging a plugin with the following structure in CI:

```
- bestImageRendererPlugin
   - executable
   - someFile.jpg
   - folder/
      - someFile.png
```

During the build, we are creating a manifest file with the list of all files & their checksums. However, when the built package is installed on an M1 Mac, the OS automagically adds `.DS_Store` files in every folder: 

```
- bestImageRendererPlugin
   - .DS_Store
   - executable
   - someFile.jpg
   - folder/
      - .DS_Store 
      - someFile.png
```

I tried creating empty `DS_Store` files in all the right places in hopes that the OS wouldn't replace them, but that did not work - they were replaced and the checksum verification failed.

I don't know why this problem occurs only on M1 macs, nor how to deal with it except for explicitly ignoring `DS_Store` files when creating & verifying the manifest. cc @AgnesToulet

ArturWierzbicki: After solving the problem described above with [this grafana PR](https://github.com/grafana/grafana/pull/46623) I ran into another one - binary signing requirement 
image

Looks like we will need to spin up a mac 😅

Proximyst:
This is now supported via Docker :)

Issue #65, opened 2019-12-04
arm64 fails to build


MarkoBursic:
If I set `make build_package ARCH=arm64` then I get error from Node.js that arm64 is unknown, but linux-arm64 is known. Then I invoked `make build_package ARCH=linux-arm64` and somewhere the arm64 got eaten and it is parsed like linux-armv6. Then it says: Not able to build for 'armv6' here, only for 'arm64'. 
OK, I understand, but how to tell him that I don't want `armv6`, rather `arm64`?

make build_package ARCH=linux-arm64
rm -rf build
./scripts/clean_target.sh linux-arm64
./node_modules/.bin/tsc
./scripts/package_target.sh linux-arm64
> pkg@4.3.8
> Fetching base Node.js binaries to PKG_CACHE_PATH
  fetched-v10.4.1-linux-armv6  [                    ] 1%
> Error! 404 Not Found
  https://github.com/zeit/pkg-fetch/releases/download/v2.5/uploaded-v2.5-node-v10.4.1-linux-armv6
> Asset not found by direct link:
  {"tag":"v2.5","name":"uploaded-v2.5-node-v10.4.1-linux-armv6"}
> Not found in GitHub releases:
  {"tag":"v2.5","name":"uploaded-v2.5-node-v10.4.1-linux-armv6"}
> Building base binary from source:
  built-v10.4.1-linux-armv6
> Error! Not able to build for 'armv6' here, only for 'arm64'
child_process.js:656
    throw err;
    ^

marefr:
See https://github.com/grafana/grafana-image-renderer/blob/master/docs/package_plugin_as_single_executable.md

We don't officially support packaging for ARM, but see https://github.com/grafana/grafana-image-renderer/issues/7#issuecomment-546599598 where someone have managed to pull it off for the remote rendering service using docker at least. 

Probably need `make build_package ARCH=linux-armd64-glibc` and you probably/maybe need to be on arm architecture to build for arm.

MarkoBursic:
Have seen all of them yet before. I do think, that your project is very good, it works on ARM on its basics without trouble, I am very impressed. There are tons of applications like IoT requiring just that what you have done here. These apps runs on cheap SBCs , mostly ARM.  Users way to 'hack' the build are very painful without any true strategy. ARM64 is a family that it will last for quite long time, IMO it is also more suitable target for your project then any other family.

marefr:
[...]

PR #332, opened 2022-03-02
Fix arm64 build for m1 with docker-desktop

gh7531:
I needed to build a Grafana image that had to run on an M1 with docker-desktop. Therefore, I tried to build using the command suggested in https://github.com/grafana/grafana-image-renderer/issues/65#issuecomment-562082850, but it failed during packaging `Unknown token 'armv6' in 'node14-linux-armv6'`.
I looked into the `package_target.sh` and saw that the `arm64` architecture was mapped to `armv6` which seems to be the issue. Changing this to `arm64` allowed me to build for `linux-arm64-glibc`.

Then I build a Grafana docker image on my m1 mac using docker-desktop and the plugin seemed to work.
image

Note:
- I only tested this on my m1 mac with docker-desktop
- I only tested the version with a bundled chromium
- I don't know if this change breaks other setups, since I don't have access to other systems.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ A Grafana backend plugin that handles rendering panels and dashboards to PNGs us
 
 ### Supported operating systems
 
-- Linux (x64)
+- Linux (x64, arm64)
 - Windows (x64)
 - Mac OS X (x64)
 
--- a/scripts/pkg.js
+++ b/scripts/pkg.js
@@ -17,8 +17,7 @@ const platformTransform = {
 const archTransform = {
   ia32: 'x84',
   arm: 'armv6',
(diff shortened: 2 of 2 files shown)


ArturWierzbicki:
hey @gh7531, thanks a lot for the contribution! Have you tried changing the `arch` in the `Makefile` to `darwin-arm64-unknown` https://github.com/grafana/grafana-image-renderer/blob/391eec1708d10730590f0ec6979bd9bbed0f09ef/Makefile#L3?

I believe this already works on the master branch for local builds, but, unfortunately, it does not work for packages intended for distribution. (https://github.com/grafana/grafana-image-renderer/pull/326/files)

PR #335, opened 2022-03-14
Set captureBeyondViewport to false by default

AgnesToulet:
Fix old alert list panel rendering (before Grafana 8 and React migration) by setting the `captureBeyondViewport` screenshot option to false by default.


spinillos (approved):
It was tested with `7.5.1` and current Grafana version and Alert list works in both with this change.

PR #338, opened 2022-03-16
Add chromium-swiftshader to enable webGL

adkinsjd:
Image rendering of panels depending on webGL (such as mapbox plots in the plotly plugin https://grafana.com/grafana/plugins/ae3e-plotly-panel/) does not work with this container. The renderer tries but fails to initialize webGL.

Adding chromium-swiftshader along with the "--use-gl=swiftshader" command line argument solves the issue.

This PR adds chromium-swiftshader to the apk package list. The command line argument can be added through the current config mechanism.


jacksongoode:
👍

PR #349, opened 2022-06-16
Add Sanitize API

ArturWierzbicki:


proto/sanitizer.proto
```diff
@@ -0,0 +1,20 @@
+syntax = "proto3";
+package pluginextensionv2;
+
+option go_package = ".;pluginextensionv2";
+
+message SanitizeRequest {
+  string filename = 1;
+  bytes content = 2;
+  bytes domPurifyConfig = 3;
+  bool allowAllLinksInSvgUseTags = 4;
```
ryantxu, line 10: maybe we have this proto definition be more "open" to non svg thigns in the future?

Perhaps 
```suggestion
  string sanitizer = 3; // 'svg' | ...
  bytes sanitizerConfig = 4; // json blob that depends on the sanitier type
```

proto/sanitizer.proto
```diff
@@ -0,0 +1,20 @@
...
+}
+
+message SanitizeResponse {
+  string error = 1;
+  bytes sanitized = 2;
```
ryantxu, line 15: is there relevant info we may also return?  I guess we can add that later

src/types.ts
```diff
@@ -25,3 +27,15 @@ export interface ImageRenderOptions extends RenderOptions {
   fullPageImage?: boolean;
   scaleImage?: number;
 }
+
+export type SanitizeRequest = {
+  filename: string;
+  content: string;
```
ryantxu, line 33: will we ever want this to support images/binary?

proto/sanitizer.proto
ArturWierzbicki: yeah - thats better. For `string sanitizer` I would probably use the library name or "sanitization logic type" instead of the file type so we avoid having an abstraction over the library config. DOMPurify has a pretty extensive API and it can support html, svg or mathml out of the box. Soo, something like this:


```
sanitizer: 'dompurify'
config: { dompurifyConfig: { ... }, domPurifyExtraFunctionalitiesBuiltByUsConfig: { ... } }
```
and

```
sanitizer: 'some-other-image-sanitizer'
config: { some-other-image-sanitizerConfig': { ... }}
```

what do you think?

src/types.ts
ArturWierzbicki: base64 ftw! :) I figured that a multipart request with `file` and `config` sections is the proper way of implementing this. back to codingg

proto/sanitizer.proto
ArturWierzbicki: not sure, I could add a generic `string info;` field and for now populate it with "sanitized" / "not sanitized" depending on whether the hash changed. We could use that field in the future if we choose to migrate to some other library that returns some info on itself.

src/types.ts
ArturWierzbicki: refactored to a multipart request

proto/sanitizer.proto
ArturWierzbicki: it would be difficult to differentiate between sanitized vs unsanitized files even right now, as dompurify modifies all svgs. I think we can do it later if we'd like

src/types.ts
```diff
@@ -25,3 +27,40 @@ export interface ImageRenderOptions extends RenderOptions {
...
+  contentType?: string;
+  domPurifyConfig?: DOMPurify.Config;
+  allowAllLinksInSvgUseTags?: boolean;
+};
```
AgnesToulet, line 37: Seems like this is never used, is there a reason to not remove it and remove the `V2` of `SanitizeRequestV2`?

src/service/http-server.ts
```diff
@@ -49,6 +61,14 @@ export class HttpServer {
 
     this.app.get('/render', asyncMiddleware(this.render));
     this.app.get('/render/csv', asyncMiddleware(this.renderCSV));
+    this.app.post(
+      '/render/sanitize',
```
AgnesToulet, line 65:
```suggestion
      '/sanitize',
```
This would be better if we could get rid of the `render` prefix and do something in Grafana to replace `/render` in the renderer URL with `/sanitize` for example.
Else, it will mess up the metrics we collect about image renderer (set up here: https://github.com/grafana/grafana-image-renderer/blob/master/src/service/metrics_middleware.ts#L49). It could be interesting to have metrics for this endpoint as well but we should keep them separate.

src/types.ts
```diff
@@ -1,3 +1,5 @@
+import * as DOMPurify from 'dompurify';
```
AgnesToulet, line 1: What about putting all these types in a `types.ts` file in `sanitizer` folder? I think it would help keep things separated

Bujupah:
hey... wondering what is the use of this feature?

ArturWierzbicki:
hey @Bujupah! The use of this feature is server side sanitization of SVGs hosted in Grafana. There is some more context here if you are interested: https://github.com/grafana/grafana/issues/50597. 
Right now, this feature is unused - we still haven't added the capability of uploading media files to Grafana.

PR #357, opened 2022-07-13
Fix invalid Content-Disposition

spinillos:
It uses `content-disposition` library to accept non-ascii characters in Content-Disposition header. Right now its failing when it's happening and we aren't creating the csv.

Related issue: https://github.com/grafana/grafana-enterprise/issues/3576

### DIFF
--- a/src/service/http-server.ts
+++ b/src/service/http-server.ts
@@ -234,7 +234,7 @@ export class HttpServer {
     const result = await this.browser.renderCSV(options);
 
     if (result.fileName) {
-      res.setHeader('Content-Disposition', `attachment; filename="${result.fileName}"`);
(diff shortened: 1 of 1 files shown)


spinillos:
Thanks Agnès!

I was more scared about breaking things and I didn't notice about variables in titles. I'm going to test it.

src/service/http-server.ts
```diff
@@ -234,7 +235,7 @@ export class HttpServer {
     const result = await this.browser.renderCSV(options);
 
     if (result.fileName) {
-      res.setHeader('Content-Disposition', `attachment; filename="${result.fileName}"`);
+      res.setHeader('Content-Disposition', contentDisposition(result.fileName));
```
ArturWierzbicki, line 238: What happens if the file name is expected to contains special characters - will they be simply removed?

ArturWierzbicki (approved):

PR #362, opened 2022-08-16
Update waiting condition for full page screenshots

spinillos:
The condition that we are using for individual panels isn't working for full page screenshots.

`panelsRendered` value is increased every time a panel is rendered. But a panel could be renderer several times when scrolling. For example, when a panel is partially visible increases `panelsRendered`, and when scroll, it increases this value again. 

So `(window as any).panelsRendered >= panelCount` could be success even if we have any panel loading data. If we have a slow query, the result could be a screenshot with a loading panel, when we are expecting to see data.

The condition for full pages is the following:
* Count `[data-panelId]` as total panels of the dashboard (rendered or not). It includes rows and panels.
* `.dashboard-row` returns the number of rows.
* `.panel-content` returns the value of RENDERED panels excluding rows (rows don't have this class).

So the condition is going to wait to all panels to render OR reach the timeout.

### DIFF
--- a/src/browser/browser.ts
+++ b/src/browser/browser.ts
@@ -361,7 +361,6 @@ export class Browser {
           height: scrollResult.scrollHeight,
         });
       }
-
       return page.screenshot({ path: options.filePath, fullPage: options.fullPageImage, captureBeyondViewport: options.fullPageImage || false });
     }, 'screenshot');
 
--- a/src/browser/browser.ts
+++ b/src/browser/browser.ts
@@ -320,10 +320,28 @@ export class Browser {
     }
 
     try {
+      await page.exposeFunction('logger', (name, value) => {
+        this.log.debug(name, value);
+      });
       await this.withTimingMetrics(() => {
         if (this.config.verboseLogging) {
           this.log.debug('Waiting for dashboard/panel to load', 'timeout', `${options.timeout}s`);
         }
+
+        if (options.fullPageImage) {
+          return page.waitForFunction(
+            () => {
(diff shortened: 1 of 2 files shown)


src/browser/browser.ts
```diff
@@ -324,14 +324,22 @@ export class Browser {
         if (this.config.verboseLogging) {
           this.log.debug('Waiting for dashboard/panel to load', 'timeout', `${options.timeout}s`);
         }
+
         return page.waitForFunction(
-          () => {
+          (isFullPage) => {
+            if (isFullPage) {
+              const panelCount = document.querySelectorAll('[data-panelId]').length;
+              const totalPanelsRendered = document.querySelectorAll('.panel-content').length + document.querySelectorAll('.dashboard-row').length;
```
ArturWierzbicki, line 332: its probably worth adding a comment what `.panel-content` refers to and why do we add the `.dashboard-row`

src/browser/browser.ts
```diff
@@ -324,14 +324,22 @@ export class Browser {
         if (this.config.verboseLogging) {
           this.log.debug('Waiting for dashboard/panel to load', 'timeout', `${options.timeout}s`);
         }
+
         return page.waitForFunction(
-          () => {
+          (isFullPage) => {
+            if (isFullPage) {
```
ArturWierzbicki, line 330: should we add a comment explaining why full page screenshots require a different waiting logic?

ArturWierzbicki (approved):
🚀

PR #365, opened 2022-08-30
Chore: Update to Node 16

Clarity-89:
- Sync dev dependency versions with Grafana and pin their exact versions
- Update Node to v 16


spinillos (approved):

xlson (approved):
Great, thanks!

PR #372, opened 2022-09-23
Update _client with _client() to avoid to fail when creating a CSV

spinillos:
It changes `_client` with `_client()` to make CSV works with last puppeteer update.

Fixes: https://github.com/grafana/grafana-image-renderer/issues

### DIFF
--- a/src/browser/browser.ts
+++ b/src/browser/browser.ts
@@ -432,7 +432,7 @@ export class Browser {
       }
     });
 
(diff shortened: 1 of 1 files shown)


Clarity-89 (approved):
Thanks for fixing this! 🎉

Bujupah:
_client is a private function and I think it is not recommended doing it that way.
I'd recommend switching to this
```ts
    const client = await page.target().createCDPSession();
    await client.send('Page.setDownloadBehavior', {
      behavior: 'allow',
      downloadPath,
    });
```

PR #375, opened 2022-09-27
Update deprecated functions

spinillos:
It updates some deprecated functions to the new ones.


Clarity-89 (approved):
Great work! 🎉  Should probably look into removing `any` types in other places as well.

PR #379, opened 2022-11-01
Handle server crash when req.failure() is null

Bujupah:
In some cases the browser request error is null and logging it directly as `req.failure().errorText` will cause the plugin to crash.

Fix is just to handle if `req.failure()` is undefined or not

### DIFF
--- a/src/browser/browser.ts
+++ b/src/browser/browser.ts
@@ -536,7 +536,7 @@ export class Browser {
   };
 
   logRequestFailed = (req: any) => {
-    this.log.error('Browser request failed', 'url', req.url(), 'method', req.method(), 'failure', req.failure().errorText);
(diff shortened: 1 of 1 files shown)


Bujupah:
Probably adding the parameter type to function will help avoiding these issues when compiling

```ts

  logRequest = (req: puppeteer.HTTPRequest) => {
    this.log.debug('Browser request', 'url', req.url(), 'method', req.method());
  };

  logRequestFailed = (req: puppeteer.HTTPRequest) => {
    this.log.error('Browser request failed', 'url', req.url(), 'method', req.method(), 'failure', req.failure()?.errorText);
  };

  logRequestFinished = (req: puppeteer.HTTPRequest) => {
    this.log.debug('Browser request finished', 'url', req.url(), 'method', req.method());
  };

```

Bujupah:
Issue is fixed in this PR #403

PR #382, opened 2022-11-07
Add delay after full page waiting time

spinillos:
It adds some extra waiting time to full page screenshots. This value is configurable.

Why? Because right now we don't have any mechanism to detect when the panel is completely rendered. The waiting time condition looks for a element that exists when the panel loaded its data and it starts to render it. But it doesn't mean that the panel is completely rendered and sometimes we take the screenshot during the rendering getting panels partially rendered.

Also we detected that loading icon and the element that indicates that the panel is ready to render could exist at the same time during this transition. So we added an extra check to guarantee that the loading icons are 0. 

We need to maintain the rest of the conditions because we could receive 0 loading icons at the moment that we open the page to render.


src/browser/browser.ts
```diff
@@ -357,6 +362,8 @@ export class Browser {
       this.log.error('Error while waiting for the panels to load', 'url', options.url, 'err', err.stack);
     }
 
+    await new Promise(r => setTimeout(r, options.fullPageImageDelayTime!! * 1000));
```
joanlopez, line 365: I think the pull request looks generally good, but this line made me realize the unit used to express the delay is explained nowhere. 

So, what do you think about slightly changing its name by `fullPageImageDelaySeconds` or at least document somewhere that the unit for that setting are seconds?

Thanks!

src/service/http-server.ts
```diff
@@ -160,6 +160,8 @@ export class HttpServer {
       timezone: req.query.timezone,
       encoding: req.query.encoding,
       deviceScaleFactor: req.query.deviceScaleFactor,
+      fullPageImage: req.query.fullPageImage,
```
AgnesToulet, line 163: What's the reason behind adding this line? I don't see this query parameter being filled on the Grafana side.

AgnesToulet:
Closing this as we didn't receive any other user feedback about this, and this may be improved with the scenes migration

PR #387, opened 2022-11-29
Added page zooming option

kaffarell:
Fixes #385
Added option for page zoom level. Allows dashboards to be zoomed (for bigger titles, etc.)

Issue #240, opened 2021-05-31
Migrate CI from CircleCI to Drone


joanlopez:
**What would you like to be added**:

Migrate the [current CircleCI configuration](https://github.com/grafana/grafana-image-renderer/tree/master/.circleci) to [Drone](https://drone.grafana.net/) by using Starlark templates to generate the YAML configuration.

The Grafana [CI configuration](https://github.com/grafana/grafana/blob/main/.drone.star) and [its Starlark scripts](https://github.com/grafana/grafana/tree/main/scripts) can be used as an example.

**Why is this needed**:

We're moving all the company CI configurations from CircleCI to Drone as the last is now the officially supported CI system, so this project should be migrated as well.

PR #390, opened 2022-12-14
Stop doing security scans on PRs

joanlopez:
**What this PR does**:

It removes the `security_scan` job from PRs' pipelines.

**Why do we need it**:

It's causing issues to / conflicting with community pull requests (like [here](https://github.com/grafana/grafana-image-renderer/pull/387#issuecomment-1330445915)) while I don't think it's that critical to keep it there - actually I don't think we continuously analyze (on a pull request basis) the output of those security scans, so I think keeping them only on master builds is enough.

In fact, that's what we do with Grafana's repo, if I'm not wrong.

**Special notes for your reviewer**:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -250,9 +250,6 @@ workflows:
     jobs:
       - build:
           filters: *filter-not-release-or-master
-      - security_scan:
-          requires:
-            - build
       - package:
           filters: *filter-not-release-or-master
           requires:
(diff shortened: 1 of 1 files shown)


jongyllen:
I agree, we can remove from PR:s

In the long run I want some scans on PR:s as well but lets tune those and think about best tooling before implementing

joanlopez:
> I agree, we can remove from PR:s
> 
> In the long run I want some scans on PR:s as well but lets tune those and think about best tooling before implementing

Absolutely! Hopefully in the long run we'll have finished #240  (as @marefr pointed out) and will much easier to tweak it depending on our needs and wishes, thaanks!!

PR #403, opened 2023-02-10
Fix error when req.failure() is null

spinillos:
Related: https://github.com/grafana/support-escalations/issues/5010

Looks like that `req.failure()` could be null and it fails when it happens.

### DIFF
--- a/src/browser/browser.ts
+++ b/src/browser/browser.ts
@@ -537,7 +537,11 @@ export class Browser {
   };
 
   logRequestFailed = (req: any) => {
-    this.log.error('Browser request failed', 'url', req.url(), 'method', req.method(), 'failure', req.failure().errorText);
+    let failureError = ""
+    if (req.failure()) {
+      failureError = req.failure().errorText
+    }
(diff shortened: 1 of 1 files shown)


src/browser/browser.ts
```diff
@@ -537,7 +537,11 @@ export class Browser {
   };
 
   logRequestFailed = (req: any) => {
-    this.log.error('Browser request failed', 'url', req.url(), 'method', req.method(), 'failure', req.failure().errorText);
+    let failureError = ""
+    if (req.failure()) {
+      failureError = req.failure().errorText
+    }
+    this.log.error('Browser request failed', 'url', req.url(), 'method', req.method(), 'failure', failureError);
   };
```
Clarity-89, line 545: Let's add some types for this:
```
  logRequestFailed = (req: puppeteer.HTTPRequest) => {
    let failureError = ""
    const failure = req?.failure();
    if (failure) {
      failureError = failure.errorText
    }
    this.log.error('Browser request failed', 'url', req.url(), 'method', req.method(), 'failure', failureError);
  };
```

Clarity-89 (approved):

PR #413, opened 2023-03-28
remove alpine edge repo

sozercan:
Adding alpine edge repo might cause dependencies to get unnecessarily bumped


sozercan:
@AgnesToulet @joanlopez can i get review on this when you can please?

xlson (approved):
Thanks for updating this. I don't think we've actually needed to depend on edge for a really long time.

Issue #388, opened 2022-11-30
Add auth_token to install instruction


xlson:
**What would you like to be added**:

Instruction on how to configure the auth_token and a link back to the grafana docs should be added to the readme that is shown in the plugin install page. See https://github.com/grafana/grafana/pull/59560

**Why is this needed**:

PR #416, opened 2023-04-11
Docs: add security section

AgnesToulet:
This is a start, I don't know if this is what you had in mind when you wrote https://github.com/grafana/grafana-image-renderer/issues/388

Closes https://github.com/grafana/grafana-image-renderer/issues/388

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,12 @@ The following example describes how to build and run the remote HTTP rendering s
 
 1. Restart Grafana.
 
+## Security
+
+You can restrict access to the rendering endpoint by specifying a secret token. The token should be configured in the Grafana configuration file and the renderer configuration file. This token is important when you run the plugin in remote rendering mode.
+
+See [Grafana Image Rendering documentation](https://grafana.com/docs/grafana/latest/image-rendering/#security) to configure this secret token.
+
 ## Configuration
 
 For available configuration settings, please refer to [Grafana Image Rendering documentation](https://grafana.com/docs/grafana/latest/image-rendering/#configuration).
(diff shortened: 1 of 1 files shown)


joanlopez (commented):
It looks like a great start, few things I'm wondering if would make sense to add are:
- Give few more details about the reasons why to set this up, like "otherwise something like this could happen", and refer the security incident.
- Perhaps mentioning that it's configured to `-` by default, so it's:
  - Not safe, because any potential attacker would start with `-` as secret token.
  - Therefore, it must be replaced.
  - But, if you get started, you'll notice that it requires a secret token to authorize rendering requests.

joanlopez:
Indeed, by looking at the code that checks for the token (for instance, changes [here](https://github.com/grafana/grafana-image-renderer/pull/417/files)), it seems it always checks for equality and empty string is not considered a valid (incoming) auth token. Therefore, perhaps we may need to slightly change the wording to say this is mandatory, at least with the default value `-`, which we strongly recommend to modify.

WDYT?

joanlopez:
Hi @AgnesToulet, is there anything else missing? 
Honestly, if we don't have any clear idea to add more details, I'd merge it as-is, and perhaps iterate it later, if needed.

joanlopez (approved):

AgnesToulet:
Sorry I missed your review on this. I updated the README according to your feedback, thanks!

joanlopez:
> Sorry I missed your review on this. I updated the README according to your feedback, thanks!

Don't worry at all! Looks great! :)

PR #417, opened 2023-04-14
Security: can set array of auth tokens

AgnesToulet:
To rotate authentication tokens in a distributed environment, updating the renderer token means breaking the rendering features in Grafana until the Grafana instances are updated with the right token.

Allowing to set up an array of authentication tokens allows to rotate tokens without downtime. 

Notes:
- In environment variables, tokens can be set as space delimited strings (as tokens could contain commas).
- Still support setting the auth token as a string to not introduce a breaking change.

### DIFF
--- a/scripts/clean_target.sh
+++ b/scripts/clean_target.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+
+ARCH="${1:-}"
+OUT="${2:-}"
+
+if [ -z "$ARCH" ]; then
+    echo "ARCH (arg 1) has to be set"
+    exit 1
+fi
+
+PLUGIN_NAME=plugin-${ARCH}
+
+if [ ! -z "$OUT"  ]; then
+    PLUGIN_NAME=${OUT}
+fi
+
+rm -rf .dist/${PLUGIN_NAME}
+rm -f ./artifacts/${PLUGIN_NAME}.zip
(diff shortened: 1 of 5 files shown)


scripts/clean_target.sh
AgnesToulet, line 1: Adding this back as I'm not sure why it was deleted as it's still used in the Makefile.

undef1nd (approved):
I have tested by setting an array of tokens in the renderer's config file, as well as via `AUTH_TOKEN` env var, and running renderer as a service. Have not noticed any issues. Would be great if someone could test it as a plugin.

src/config.ts
```diff
@@ -148,3 +148,12 @@ export const readJSONFileSync = (filePath: string): any => {
   const rawdata = fs.readFileSync(filePath, 'utf8');
   return JSON.parse(rawdata);
 };
+
+export const isAuthTokenValid = (config: SecurityConfig, reqAuthToken: string): boolean => {
+  let configToken = config.authToken || [''];
+  if (typeof configToken == "string") {
```
spinillos, line 154:
```suggestion
  if (typeof configToken === "string") {
```

Shouldn't be `===`?

spinillos (approved):
Tested by HTTP and Plugin and its working well

joanlopez (approved):

Issue #432, opened 2023-06-13
Enhancement: Bump puppeteer to switch to using "Chrome for Testing"


tolzhabayev:
**What would you like to be added**:
Since `20.1.x` puppeteer is switching from "Chromium" to "[Chrome for Testing](https://developer.chrome.com/en/blog/chrome-for-testing/)" by [default](https://github.com/puppeteer/puppeteer/pull/10054)

So I would consider bumping the library and see how it improved the stability of the image renderer.

**Why is this needed**:
- Better reliability and stability

PR #433, opened 2023-06-13
Puppeteer: upgrade to v21

Clarity-89:
Upgrade Puppeteer to the latest version to take advantage of the new features. 

Fixes https://github.com/grafana/grafana-image-renderer/issues/432


spinillos (approved):
I tested with some dashboards.

Clarity-89:
> I tested with some dashboards.

It works locally, but the build fails as some dependencies need to be updated there as well.

joanlopez:
> It works locally, but the build fails as some dependencies need to be updated there as well.

Do we know exactly which dependencies need to be updated? 🤔

Clarity-89:
> > It works locally, but the build fails as some dependencies need to be updated there as well.
> 
> Do we know exactly which dependencies need to be updated? 🤔

We're relying on some [internals](https://github.com/grafana/grafana-image-renderer/blob/master/scripts/download_chromium.js#L4) when downloading Chronium, which have been removed in the latest versions. Need to investigate the alternatives.

package.json
```diff
@@ -76,7 +78,8 @@
     ]
   },
   "pkg": {
-    "assets": "proto/*"
+    "assets": "proto/*",
+    "scripts": "node_modules/vm2/lib/setup-sandbox.js"
```
AgnesToulet, line 82: Why do we need this? 

This library is deprecated and, if we really need it, we should use `isolated-vm`:
image

scripts/download_chromium.js
```diff
@@ -1,48 +1,46 @@
...
 
 if (platform === 'win32' && arch === 'x64') {
-  platform = 'win64'
+    platform = BrowserPlatform.WIN64;
 }
 
 if (platform === 'darwin') {
-  platform = 'mac'
+    platform = BrowserPlatform.MAC;
 }
```
AgnesToulet, line 23: I think this can be removed, in the `@puppeteer/browsers` doc, it's said that it'auto-detected: https://pptr.dev/browsers-api/browsers.installoptions#:~:text=platform,Auto%2Ddetected.

scripts/download_chromium.js
AgnesToulet, line 1: I think this should be rename to `download_chrome.js`.

scripts/package_target.sh
```diff
@@ -16,7 +16,7 @@ if [ $? != 0 ]; then
 fi
 
 if [ ${SKIP_CHROMIUM} = false ]; then
-    node scripts/download_chromium.js ${ARCH} ${OUT}
+   node scripts/download_chromium.js ${ARCH} ${OUT}
```
AgnesToulet, line 19:
```suggestion
   node scripts/download_chrome.js ${ARCH} ${OUT}
```
If you agree with my previous suggestion.

scripts/package_target.sh
```diff
@@ -16,7 +16,7 @@ if [ $? != 0 ]; then
...
 else
     echo "Skipping chromium download"
```
AgnesToulet, line 21:
```suggestion
    echo "Skipping chrome download"
```
Same.

package.json
Clarity-89: Apparently this is required for the latest puppeteer to work. See [here](https://github.com/vercel/pkg/issues/1947). I'll look into replacing it with isolated-vim. Afaik it's only necessary during build, so can be also moved to dev dependencies.

scripts/download_chromium.js
Clarity-89: But what if we explicitly need to build binaries for specific arch? E.g. I'm on arm64 and want to build and package the plugin for Windows?

scripts/download_chromium.js
Clarity-89: I was actually thinking that maybe we'd just stick to chromium, to be consistent with Docker images?

package.json
Clarity-89: It looks like this dependency [has been removed](https://github.com/puppeteer/puppeteer/releases/tag/browsers-v1.4.5) in the latest version of `@puppeteer/browsers`, so it's not necessary anymore.

scripts/download_chromium.js
Clarity-89: I switched the download back to chromium for consistency.

docker-compose.yml
```diff
@@ -0,0 +1,17 @@
+version: '2'
```
joanlopez, line 1: Are you willing to merge this file? If so, I have few questions:
- I don't see any `./data/license.jwt`, so I guess we should leave some instructions, at least.
- Where's the `renderer` image name taken from? The `Makefile` refers to `grafana/grafana-image-renderer:dev`, so I guess we should either change it accordingly, or point to the Dockerfile from the `docker-compose.yaml`.

yarn.lock
```diff
@@ -2,6 +2,13 @@
 # yarn lockfile v1
```
joanlopez, line 2: Is this a good opportunity to switch to v2? Perhaps now, but in a separate PR?

What would be the recommendation here to work locally? 
Have both, Yarn v2+ for Grafana and Yarn v1 for this project?

Thanks!

docker-compose.yml
Clarity-89: My bad, used this for local testing and committed it accidentally, will remove.

yarn.lock
Clarity-89: Not sure what would be best here, I'll investigate switching to V2 on a separate PR.

joanlopez: Great, thanks!

PS: Honestly, if there's any easy way to have/use both with no big pains, I'd be happy to keep it as-is, although it's something we'll need to revisit anyway at some time, yeah.

AgnesToulet:
I tested this in plugin mode on Windows and there was one issue with chromium path.
The line https://github.com/grafana/grafana-image-renderer/blob/master/src/app.ts#L12 should be updated to:
```
const chromeFolderPrefix = 'chromium'
```

Apart from that, it seems to work fine 👍 

If we don't use Chrome in this PR, I think we should do it in a follow-up one but also look into the performance to ensure it isn't worse (it shouldn't but we should check). I would leave https://github.com/grafana/grafana-image-renderer/issues/432 open to keep track of that.

AgnesToulet:
It seems like installing Chrome on Alpine is not recommended so we should keep Chromium here and switch to chrome-stable in the [Debian image](https://github.com/grafana/grafana-image-renderer/blob/master/debian.Dockerfile). This is the Puppeteer Dockerfile: https://github.com/puppeteer/puppeteer/blob/main/docker/Dockerfile

If the performance is much better with Chrome for Testing than with chromium we should update the release pipeline to also publish the Debian Docker image. 

@xlson I think you have a better knowledge around Docker than us, any opinion on the above?

xlson:
> 

I'm not following, what is it you want opinions on?

AgnesToulet:
@xlson I'd like your opinion on this:
> It seems like installing Chrome on Alpine is not recommended so we should keep Chromium here and switch to chrome-stable in the [Debian image](https://github.com/grafana/grafana-image-renderer/blob/master/debian.Dockerfile). This is the Puppeteer Dockerfile: https://github.com/puppeteer/puppeteer/blob/main/docker/Dockerfile
> 
> If the performance is much better with Chrome for Testing than with chromium we should update the release pipeline to also publish the Debian Docker image.

The context is that Puppeteer (the library we use to open a browser and take a Grafana screenshot) is switching from Chromium to Chrome for Testing by default. We'd like to upgrade to do the same if this improves the performance and stability of the image renderer but we have trouble updating the Docker images (well mostly the Alpine one that is the one we currently publish to DockerHub).

Clarity-89:
> If the performance is much better with Chrome for Testing than with chromium we should update the release pipeline to also publish the Debian Docker image.
> 

There's also a request for Debian images: https://github.com/grafana/grafana-image-renderer/issues/442. Don't see why we shouldn't publish those.

joanlopez:
> > If the performance is much better with Chrome for Testing than with chromium we should update the release pipeline to also publish the Debian Docker image.
> 
> There's also a request for Debian images: #442. Don't see why we shouldn't publish those.

I agree, it would be nice to publish both.

What I'm wondering now is: should we use a different Chrome flavour on each? If so, how much would that increase the maintainability cost?

cc/ @xlson @AgnesToulet

xlson:
If the performance is better with Chrome I think it makes sense to switch to debian. I'm sceptical of publishing both alpine and debian though as that will increase our maintenance cost.

Clarity-89:
@xlson @AgnesToulet are there any risks with switching to Debian? If not I think we'd do it.

Clarity-89:
It looks like we already have a [Debian Docker image](https://github.com/grafana/grafana-image-renderer/blob/master/debian.Dockerfile), but we just don't publish it.

xlson:
I don't really think so. We'd rather use alpine as it's designed for docker, but debian should work fine as well.

dev.json
```diff
@@ -1,6 +1,6 @@
 {
   "service": {
-    "host": "localhost",
+    "host": "0.0.0.0",
```
joanlopez, line 3: Nit; I guess it's not super important, but I'm not sure how fair is to set it like this. I've heard some complains from security folks for using `0.0.0.0` (_by default_) as it would listen for any incoming connection. Up to you!

joanlopez (approved):
Overall looks good!! 👍🏻

dev.json
AgnesToulet: Is there any reason to make this change?

Clarity-89: Without this I wasn't able to connect to the renderer when running it inside Docker against a locally run Grafana.

Clarity-89: Now I'm actually wondering why is this file used by docker, as it should be only applied for development.

AgnesToulet: Hmm I don't have this issue when I run the image renderer Docker image. How do you run it exactly?

Clarity-89: Not sure what happened, but I tried again and it works with localhost 🙈 I'll revert this change.

Clarity-89: Not sure what happened, but I tried again and it works with localhost 🙈 I'll revert this change.

AgnesToulet:
So I think this remains:
> I tested this in plugin mode on Windows and there was one issue with chromium path.
The line https://github.com/grafana/grafana-image-renderer/blob/master/src/app.ts#L12 should be updated to:
>
> const chromeFolderPrefix = 'chromium'

And we should be good to merge!

Clarity-89:
> 

Ah, that's why it wasn't working for me locally, fixed now!

AgnesToulet (approved):
I did a few more tests and everything works fine! Great work on this 👍

Issue #449, opened 2023-08-25
Plugin v3.8.0 does not load: "Could not find Chrome (ver. 116.0.5845.96)" from puppeteer


chennin:
**What happened**:

Image renderer plugin version 3.8.0 does not load, giving logs:

```
logger=plugin.grafana-image-renderer t=2023-08-25T21:15:31.654884325Z level=debug msg="Error: Could not find Chrome (ver. 116.0.5845.96). This can occur if either"
logger=plugin.grafana-image-renderer t=2023-08-25T21:15:31.654903285Z level=debug msg=" 1. you did not perform an installation before running the script (e.g. `npm install`) or"
logger=plugin.grafana-image-renderer t=2023-08-25T21:15:31.654910725Z level=debug msg=" 2. your cache path is incorrectly configured (which is: /usr/share/grafana/.cache/puppeteer)."
logger=plugin.grafana-image-renderer t=2023-08-25T21:15:31.654914485Z level=debug msg="For (2), check out our guide on configuring puppeteer at https://pptr.dev/guides/configuration."
logger=plugin.grafana-image-renderer t=2023-08-25T21:15:31.654917835Z level=debug msg="    at ChromeLauncher.resolveExecutablePath (/snapshot/src/node_modules/puppeteer-core/lib/cjs/puppeteer/node/ProductLauncher.js:300:27)"
logger=plugin.grafana-image-renderer t=2023-08-25T21:15:31.654922826Z level=debug msg="    at ChromeLauncher.executablePath (/snapshot/src/node_modules/puppeteer-core/lib/cjs/puppeteer/node/ChromeLauncher.js:180:25)"
logger=plugin.grafana-image-renderer t=2023-08-25T21:15:31.654926176Z level=debug msg="    at PuppeteerNode.executablePath (/snapshot/src/node_modules/puppeteer-core/lib/cjs/puppeteer/node/PuppeteerNode.js:170:31)"
logger=plugin.grafana-image-renderer t=2023-08-25T21:15:31.654929406Z level=debug msg="    at main (/snapshot/src/build/app.js:24:46)"
logger=plugin.grafana-image-renderer t=2023-08-25T21:15:31.654932416Z level=debug msg="    at Object.<anonymous> (/snapshot/src/build/app.js:75:1)"
logger=plugin.grafana-image-renderer t=2023-08-25T21:15:31.654935376Z level=debug msg="    at Module._compile (pkg/prelude/bootstrap.js:1926:22)"
[...]

PR #451, opened 2023-09-01
Plugin: fix Chrome path

AgnesToulet:
The Puppeteer update (https://github.com/grafana/grafana-image-renderer/pull/433) contained an update to the `puppeteer.executablePath` function used in plugin mode to compute Chrome executable path. 

This PR updates the package scripts to ensure Chrome is always installed under `<pluginDir>/chrome` so we can use this path in plugin mode. 

Fixes https://github.com/grafana/grafana-image-renderer/issues/449

### DIFF
--- a/scripts/download_chrome.js
+++ b/scripts/download_chrome.js
@@ -1,4 +1,5 @@
 const { BrowserPlatform, Browser, install, resolveBuildId } = require('@puppeteer/browsers');
+const childProcess = require('child_process');
 const path = require('path');
 
 const archArg = process.argv[2];
@@ -25,6 +26,9 @@ const outputPath = path.resolve(process.cwd(), 'dist', process.argv[3] || `plugi
 
 const browserVersion = Browser.CHROME;
 
+// Clean up download folder if exists
+childProcess.execSync(`rm -rf "${outputPath}/chrome"`);
+
 async function download() {
     const buildId = await resolveBuildId(browserVersion, platform, 'latest');
     console.log(`Installing ${browserVersion} into ${outputPath}`);
@@ -38,4 +42,14 @@ async function download() {
 
 download().then(() => {
     console.log(`${browserVersion} downloaded into:`, outputPath);
+
+    // All the chrome files should be in the plugin 'chrome' folder
+    // This is used to set the puppeteer.executablePath
+    const out = childProcess.execSync(`find ${outputPath}/chrome -type f -name chrome.exe -exec dirname "{}" \;`);
+    const chromeBinDir = out.toString().trim()
+    
+    console.log(`Moving ${chromeBinDir} content into ${outputPath}/chrome`)
+    childProcess.execSync(`mv ${chromeBinDir} ${outputPath}/tmp`)
+    childProcess.execSync(`rm -r "${outputPath}/chrome"`);
+    childProcess.execSync(`mv "${outputPath}/tmp" "${outputPath}/chrome"`);
 });
--- a/scripts/download_chrome.js
+++ b/scripts/download_chrome.js
@@ -1,5 +1,4 @@
 const { BrowserPlatform, Browser, install, resolveBuildId } = require('@puppeteer/browsers');
-const childProcess = require('child_process');
 const fs = require('fs')
 const path = require('path');
 
@@ -26,9 +25,6 @@ if (platform === 'darwin') {
 const outputPath = path.resolve(process.cwd(), 'dist', process.argv[3] || `plugin-${archArg}`);
 
 const browserVersion = Browser.CHROME;
-
-// Clean up download folder if exists
-childProcess.execFileSync('rm', ['-rf', `${outputPath}/chrome`]);
 
 async function download() {
     const buildId = await resolveBuildId(browserVersion, platform, 'latest');
--- a/scripts/download_chrome.js
+++ b/scripts/download_chrome.js
@@ -45,7 +45,8 @@ download().then(() => {
 
     // All the chrome files should be in the plugin 'chrome' folder
     // This is used to set the puppeteer.executablePath
-    const out = childProcess.execFileSync('find', [`${outputPath}/chrome`, '-type', 'f', '-name', 'chrome.exe', '-exec', 'dirname', '{}', '\;']);
+    const ext = platform === BrowserPlatform.WIN64 ? '.exe' : '';
+    const out = childProcess.execFileSync('find', [`${outputPath}/chrome`, '-type', 'f', '-name', `chrome${ext}`, '-exec', 'dirname', '{}', '\;']);
     const chromeBinDir = out.toString().trim()
     
     console.log(`Moving ${chromeBinDir} content into ${outputPath}/chrome`);
--- a/scripts/download_chrome.js
+++ b/scripts/download_chrome.js
@@ -1,5 +1,6 @@
 const { BrowserPlatform, Browser, install, resolveBuildId } = require('@puppeteer/browsers');
 const childProcess = require('child_process');
+const fs = require('fs')
 const path = require('path');
 
 const archArg = process.argv[2];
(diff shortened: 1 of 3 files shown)


scripts/download_chrome.js
```diff
@@ -38,4 +42,15 @@ async function download() {
 
 download().then(() => {
     console.log(`${browserVersion} downloaded into:`, outputPath);
+
+    // All the chrome files should be in the plugin 'chrome' folder
+    // This is used to set the puppeteer.executablePath
+    const ext = platform === BrowserPlatform.WIN64 ? '.exe' : '';
+    const out = childProcess.execFileSync('find', [`${outputPath}/chrome`, '-type', 'f', '-name', `chrome${ext}`, '-exec', 'dirname', '{}', '\;']);
```
Clarity-89, line 49: This returns an empty string for me and as a result, the whole build fails. Also I'm not sure I get the purposes of this, before the changes the chrome was being downloaded into `${outputPath}/chrome`, which is the built plugin's folder.

scripts/download_chrome.js
```diff
@@ -38,4 +42,15 @@ async function download() {
...
+    const chromeBinDir = out.toString().trim()
+    
+    console.log(`Moving ${chromeBinDir} content into ${outputPath}/chrome`);
+    childProcess.execFileSync('mv', [chromeBinDir, `${outputPath}/tmp`]);
```
Clarity-89, line 53: This command fails when there's no `${outputPath}/tmp` directory.

scripts/download_chrome.js
AgnesToulet: The thing is in `app.ts` we need to know where `chrome.exe` is and currently it's in `${outputPath}/chrome/<platform>-<buildVersion>/chrome-<patform>`. The goal of this PR is to get rid of these two intermediate folders that are platform/build-specific so we have a generic path that can be used everywhere. 

Before, we used `executablePath()` to get this information but now that it's not working anymore, I couldn't find another way to do this.

AgnesToulet: The folder structure or the file name may be different on Mac, I'll take a look.

Clarity-89: Ah ok, I see, makes sense.

Clarity-89: Actually, I think it's [this one](https://github.com/puppeteer/puppeteer/blob/main/packages/browsers/src/launch.ts#L64).

scripts/download_chrome.js
```diff
@@ -25,6 +26,9 @@ const outputPath = path.resolve(process.cwd(), 'dist', process.argv[3] || `plugi
 
 const browserVersion = Browser.CHROME;
 
+// Clean up download folder if exists
+childProcess.execFileSync('rm', ['-rf', `${outputPath}/chrome`]);
+
 async function download() {
```
Clarity-89, line 32:
```js
async function download() {
    const buildId = await resolveBuildId(browserVersion, platform, 'latest');
    console.log(`Installing ${browserVersion} into ${outputPath}`);
    const executablePath = computeExecutablePath({buildId, browser: browserVersion, platform, cacheDir: ''});
    const browser = await install({
        cacheDir: outputPath,
        browser: browserVersion,
        platform,
        buildId,
    });

    return {browser, executablePath};
}

download().then(({executablePath}) => {
    console.log(`${browserVersion} downloaded into:`, outputPath);

    // All the chrome files should be in the plugin 'chrome' folder
    // This is used to set the puppeteer.executablePath
    const out = executablePath.substring(0, executablePath.lastIndexOf("/"));
    const chromeBinDir = `${outputPath}/${out}`
    console.log(`Moving ${chromeBinDir} content into ${outputPath}/chrome`);
    childProcess.execFileSync('mv', [chromeBinDir, `${outputPath}/tmp`]);
    childProcess.execFileSync('rm', ['-r', `${outputPath}/chrome`]);
    childProcess.execFileSync('mv', [`${outputPath}/tmp`, `${outputPath}/chrome`]);
});
```

This should work independent of the OS and move the chrome executable directly to the `chrome` folder. The path also needs to be adjusted in `app.ts`. However, this doesn't fix the plugin startup error for me, not sure if it'd be different on Windows.

AgnesToulet: I think this doesn't work on Mac because Chrome executable is not at the root of the Chrome folder as on Linux or Windows. 

I'll look into another solution. I think if we can pass the `buildId` to the package, then we can use the `computeExecutableFunction` to get the whole executable path.

src/app.ts
```diff
@@ -78,7 +75,13 @@ async function main() {
...
 
 main().catch((err) => {
-  console.error(err);
+  const errorLog = {
+    '@level': 'error',
+    '@message': 'failed to start grafana-image-renderer',
+    'error': err.message,
+    'trace': err.stack,
+  }
+  console.error(JSON.stringify(errorLog));
```
AgnesToulet, line 84: This error was actually logged as a "debug" log in plugin mode so that's why it was usually not displayed. This is fixing that and it should help us getting more info from user issues.

AgnesToulet:
@Clarity-89 I found another solution, this should be more portable and I find it much cleaner than the other one. Tested on Windows only, I'll do some tests on Linux.

Clarity-89:
> @Clarity-89 I found another solution, this should be more portable and I find it much cleaner than the other one. Tested on Windows only, I'll do some tests on Linux.

Tested on Mac and this works! 🎉  I wonder if this still going to work in docker, since we're still using Chromium in our image.

Clarity-89:
Ok, yeah the Docker image built locally works for me. I'm asking because according to [this issue](https://github.com/grafana/grafana-image-renderer/issues/449) the renderer also fails in Docker on Linux. Is that a separate issue?

src/app.ts
```diff
@@ -21,20 +20,18 @@ async function main() {
...
-        }
-
-        config.rendering.chromeBin = [path.dirname(process.execPath), ...parts].join(path.sep);
-      } else {
-        // local chrome installation in dev mode
-        config.rendering.chromeBin = env['PUPPETEER_EXECUTABLE_PATH'];
-      }
+      const execPath = path.dirname(process.execPath);
+      const chromeInfoFile = fs.readFileSync(path.resolve(execPath, 'chrome-info.json'), 'utf8');
+      const chromeInfo = JSON.parse(chromeInfoFile);
```
Clarity-89, line 25: We'd probably handle the possible errors here and maybe default to executable path from puppeteer (the one in the `.cache` folder)?

src/app.ts
```diff
@@ -21,20 +20,18 @@ async function main() {
...
+
+      const executablePath = computeExecutablePath({
+        cacheDir: path.dirname(process.execPath),
+        browser: Browser.CHROME,
+        buildId: chromeInfo.buildId,
+      });
+
+      config.rendering.chromeBin = executablePath;
```
Clarity-89, line 33:
```suggestion
     config.rendering.chromeBin = computeExecutablePath({
        cacheDir: path.dirname(process.execPath),
        browser: Browser.CHROME,
        buildId: chromeInfo.buildId,
      });
```

NIT: Don't think an extra variable is necessary here.

AgnesToulet:
> I'm asking because according to https://github.com/grafana/grafana-image-renderer/issues/449 the renderer also fails in Docker on Linux. Is that a separate issue?

Yes it's a bit different but also fixed in this PR. In this custom Docker image, the executable path is provided through the `GF_PLUGIN_RENDERING_CHROME_BIN` env variable. But as this variable was used after the executable path computation, it also failed on the call to the `executablePath()` function. 
This is why I moved this section:
```
  if (env['GF_PLUGIN_RENDERING_CHROME_BIN']) {
    config.rendering.chromeBin = env['GF_PLUGIN_RENDERING_CHROME_BIN'];
  }
```
We shouldn't compute the executable path anymore if it's provided.

Clarity-89 (approved):
Great work on this! 🎉

PR #469, opened 2023-10-05
Drone: Fix push to Docker step

AgnesToulet:
Switch all scripts execution to be executed with /bin/bash instead of sh. This fixes the [Drone current failure](https://drone.grafana.net/grafana/grafana-image-renderer/194/1/10) as the sh shell has no syntax to create arrays, but Bash has.

### DIFF
--- a/scripts/build_push_docker.sh
+++ b/scripts/build_push_docker.sh
@@ -12,10 +12,9 @@ fi
 
 echo "building ${TAG}"
 echo "$DOCKER_PASS" | docker login -u "$DOCKER_USER" --password-stdin
-tags=()
-tags+=${IMAGE_NAME}:${TAG}
+tags=("-t ${IMAGE_NAME}:${TAG}")
 if [ -z "$(echo $TAG | grep -E "beta|master")" ]; then
-  tags+={IMAGE_NAME}:latest
+  tags+=("-t ${IMAGE_NAME}:latest")
 fi
 
-# docker buildx build --platform linux/amd64,linux/arm64 --push -t ${tags[@]} .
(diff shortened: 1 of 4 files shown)


scripts/build_push_docker.sh
```diff
@@ -18,4 +18,4 @@ if [ -z "$(echo $TAG | grep -E "beta|master")" ]; then
   tags+={IMAGE_NAME}:latest
 fi
 
-docker buildx build --platform linux/amd64,linux/arm64 --push -t ${tags[@]} .
+# docker buildx build --platform linux/amd64,linux/arm64 --push -t ${tags[@]} .
```
joanlopez, line 21: Is this intentional? 🤔

michbeck100, line 21: This should be reverted. Apart from that it won't work because you have to use `-t` for every tag.

scripts/build_push_docker.sh
michbeck100, line 1: Please see [this commit](https://github.com/grafana/grafana-image-renderer/compare/master...michbeck100:grafana-image-renderer:fix-docker-build) for additional changes. This will create multiple `-t` tags.

AgnesToulet: Thanks, I updated it.

joanlopez (approved):
LGTM!

michbeck100:
Seems that the docker command doesn't support `--platform`: https://drone.grafana.net/grafana/grafana-image-renderer/197/1/10

Any ideas?

michbeck100:
This should fix it: https://github.com/grafana/grafana-image-renderer/pull/470

The image used to run the docker build was too old to include buildx

PR #470, opened 2023-10-05
update google/cloud-sdk image

michbeck100:


AgnesToulet:
Thanks! I created a separate PR as the new Drone file needs to be signed. I also spotted another issue that should be fixed by https://github.com/grafana/grafana-image-renderer/pull/471.

PR #472, opened 2023-10-05
Browser: Fix panel rendered waiting condition

AgnesToulet:
Fix the following things in the panel rendered waiting condition:
- `panelCount` was always 0, now it should be the right count for valid panels and 0 for invalid panels (e.g. panel not found error)
- `panelsRendered` is always undefined at first and is set when the panel is rendered so checking if it's undefined make the waiting condition pass on the first check and we actually don't wait for the panel to be rendered

### DIFF
--- a/src/browser/browser.ts
+++ b/src/browser/browser.ts
@@ -47,7 +47,7 @@ export class Browser {
     }
   }
 
-  async start(): Promise<void> {}
+  async start(): Promise<void> { }
 
   validateRenderOptions(options: RenderOptions) {
     if (options.url.startsWith(`socket://`)) {
@@ -216,7 +216,7 @@ export class Browser {
         heights.dashboard.client,
         scrollDivSelector
       );
-      
+
       await new Promise(executor => setTimeout(executor, scrollDelay));
     }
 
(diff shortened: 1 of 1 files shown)


src/browser/browser.ts
```diff
@@ -351,8 +351,8 @@ export class Browser {
               return totalPanelsRendered >= panelCount;
             }
 
-            const panelCount = document.querySelectorAll('.panel').length || document.querySelectorAll('.panel-container').length;
-            return (window as any).panelsRendered >= panelCount || (window as any).panelsRendered === undefined;
+            const panelCount = document.querySelectorAll('.panel-solo').length || document.querySelectorAll('[class$=\'panel-container\']').length;
```
juanicabanas, line 354: why did it change? Does it work right now?

juanicabanas (approved):

src/browser/browser.ts
AgnesToulet: I think it changes a while ago in Grafana but we didn't notice because it was 0 and `panelsRendered` is `undefined` until the panel is rendered. So for one panel (what we always render because the condition for full page is different), we checked `undefined >= 0` for a while (which is false) and when the panel is rendered, we checked `0 >= 0` (which is true) so it returned. 

I don't think we actually need this and we could set it to 1 but I would need to think more about it before making this more bigger change (and it would be nice to have a test suite as well for this kind of things).

PR #487, opened 2023-12-17
Support PDF rendering

ryantxu:
See also: https://github.com/grafana/grafana/pull/79612

http://localhost:3000/render/d/MP-Di9F7k/candlestick?orgId=1&kiosk=tv&from=1626214470464&to=1626216259472&width=1200&theme=light&encoding=pdf
image

http://localhost:3000/render/d/MP-Di9F7k/candlestick?orgId=1&kiosk=tv&from=1626214470464&to=1626216259472&width=1200&theme=dark&encoding=pdf
image

This does not change anything that would make rendering work better or look nice, just letting pdf files be pdfs rather than png files


src/browser/pdf.ts
```diff
@@ -0,0 +1,15 @@
...
+  return {
+    landscape: urlParams.get('pdf.landscape') === 'true',
+    format: (urlParams.get('pdf.format') as puppeteer.PaperFormat) ?? 'Letter',
+    preferCSSPageSize: urlParams.get('pdf.preferCSSPageSize') !== 'false', // defaults true
+    omitBackground: urlParams.get('pdf.omitBackground') !== 'false', // defaults true,
+    printBackground: urlParams.get('pdf.printBackground') === 'true', 
+    pageRanges: urlParams.get('pdf.pageRanges') ?? undefined,
+    tagged: urlParams.get('pdf.tagged') === 'true',
+  };
+}
```
ryantxu, line 15: This allows for changing properties via URL eg:

http://localhost:3000/render/d/MP-Di9F7k/candlestick?orgId=1&kiosk=tv&from=1626214470464&to=1626216259472&theme=light&encoding=pdf&pdf.landscape=true&pdf.format=Legal

src/browser/browser.ts
```diff
@@ -381,10 +383,18 @@ export class Browser {
...
       }
+
+      if (isPDF) {
+        return page.pdf({
+          ...getPDFOptionsFromURL(options.url),
+          path: options.filePath,
+        })
+      }
+
       return page.screenshot({ path: options.filePath, fullPage: options.fullPageImage, captureBeyondViewport: options.fullPageImage || false });
```
torkelo, line 394: should we wrap this in if(!isPDF ) ?

ryantxu: not necessary given the `if(isPDF) ...` case returns, but perhaps an else clause would be cleaner (I don't have any strong opinions here)

src/plugin/v2/grpc_plugin.ts
```diff
@@ -137,6 +137,11 @@ class PluginGRPCServer {
       headers: headers,
     };
 
+    // Encoding parameter is not send over grpc
+    if (req.url.includes("encoding=pdf")) {
+      options.encoding = "pdf"
+    }
```
AgnesToulet, line 143: What's the benefit of doing this compared to adding this field in the GRPC request object?

ryantxu: I would flip the question -- what is the benefit of encoding it in gRPC?

From what I see, we are sending the encoding parameter in the URL regardless -- but honestly unclear how it is used in other contexts

ryantxu: the main downside I see is that we would also have to send our format in the http flavor -- I suppose next to encoding?

AgnesToulet: First, having `encoding` as a gRPC parameter would be more consistent with the server mode that already reads the `encoding` option from the request parameters, not the URL ones. 
It's also a cleaner solution as this will be clearly defined as a parameter in the proto files, rather than in the URL where nothing is defined/documented. Parameters in the URL should only be used by Grafana when the image renderer is accessing the URL through Chrome. I understand your choice for the PDF options, and I think it's OK to let them live in the URL as Grafana will also use them, but I don't think it's the case for `encoding`. 

Also, you did your tests with the `/render` endpoint but the goal is to use this in the reporting feature where it won't be in the URL.

AgnesToulet:
Thanks for your work on this feature! I opened a PR to make some changes to it: https://github.com/grafana/grafana-image-renderer/pull/495

I'm working on updates to the reporting feature that will rely on that. I'll open the associated PRs soon.

Bujupah:
I was working on improving this render and I worked on something similar to this...
The pdf enhancement I did have also some rebranding and cover page but bit limited to simple layout only...

If ever interested take a look :p
https://github.com/Bujupah/yet-another-renderer/issues/12#issuecomment-2016263550

PR #496, opened 2024-02-11
WaitingForPanels: Change waiting logic

torkelo:
This improves and simplifies the waiting for panel queries and panel rendering logic. 

Instead of scanning for how many panels there are using css selectors and comparing this to panel content (which is a bit unreliable / fragile) we just wait for all network requests to be completed and then an additional 100ms for data rendering (this step should be blocking so maybe we can lower this). 

* [x] Now the waiting for panels step also works in scenes
* [x] Tested with single panels
* [x] Tested with full dashboard images (using height=-1)  (With many slow queries)
* [x] Tested with new and angular panels

### DIFF
--- a/src/browser/browser.ts
+++ b/src/browser/browser.ts
@@ -321,46 +321,15 @@ export class Browser {
     }
 
     const isPDF = options.encoding === 'pdf';
+
     try {
-      await this.withTimingMetrics(() => {
+      await this.withTimingMetrics(async () => {
         if (this.config.verboseLogging) {
           this.log.debug('Waiting for dashboard/panel to load', 'timeout', `${options.timeout}s`);
         }
 
-        return page.waitForFunction(
-          (isFullPage) => {
-            /**
-             * panelsRendered value is updated every time that a panel renders. It could happen multiple times in a same panel because scrolling. For full page screenshots
-             * we can reach panelsRendered >= panelCount condition even if we have panels that are still loading data and their panelsRenderer value is 0, generating
-             * a screenshot with loading panels. It's why the condition for full pages is different from a single panel.
-             */
-            if (isFullPage) {
-              /**
-               * data-panelId is the total number of the panels in the dashboard. Rows included.
-               * panel-content only exists in non-row panels when the data is loaded.
-               * dashboard-row exists only in rows.
-               */
-              const panelCount = document.querySelectorAll('[data-panelId]').length;
-              const panelsRendered = document.querySelectorAll("[class$='panel-content']");
-              let panelsRenderedCount = 0;
-              panelsRendered.forEach((value: Element) => {
-                if (value.childElementCount > 0) {
-                  panelsRenderedCount++;
-                }
-              });
-
-              const totalPanelsRendered = panelsRenderedCount + document.querySelectorAll('.dashboard-row').length;
-              return totalPanelsRendered >= panelCount;
-            }
-
-            const panelCount = document.querySelectorAll('.panel-solo').length || document.querySelectorAll("[class$='panel-container']").length;
-            return (window as any).panelsRendered >= panelCount || panelCount === 0;
-          },
-          {
-            timeout: options.timeout * 1000,
-          },
-          options.fullPageImage || isPDF
-        );
+        await page.waitForNetworkIdle();
+        await new Promise((resolve) => setTimeout(resolve, 100));
       }, 'panelsRendered');
     } catch (err) {
       this.log.error('Error while waiting for the panels to load', 'url', options.url, 'err', err.stack);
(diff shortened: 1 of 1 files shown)


src/browser/browser.ts
```diff
@@ -321,46 +321,15 @@ export class Browser {
...
-
-            const panelCount = document.querySelectorAll('.panel-solo').length || document.querySelectorAll("[class$='panel-container']").length;
-            return (window as any).panelsRendered >= panelCount || panelCount === 0;
-          },
-          {
-            timeout: options.timeout * 1000,
-          },
-          options.fullPageImage || isPDF
-        );
+        await page.waitForNetworkIdle({ timeout: options.timeout * 1000 });
```
AgnesToulet, line 331: I think this is the same as what we do here: https://github.com/grafana/grafana-image-renderer/blob/eae11cd4999759705bc9fbf1054873f7b5868479/src/browser/browser.ts#L303

AgnesToulet (commented):
I think this doesn't cover async queries (I discovered them recently and I'd say this is the only reason why we need this complicated code). An example of data source using them is the Athena DS. 

> Now the waiting for panels step also works in scenes

I don't understand this. The new PDF feature relying on scenes is working fine with slow and normal queries. The single panel rendering is broken but it doesn't seem related to timeout (I'll take a look at that by the way).

src/browser/browser.ts
torkelo: @AgnesToulet  not 100% sure this is the same, this is just for the initial page load, before we scroll down and trigger more queries. That is why I do a second call to waitForNetworkIdle here as this is after the scroll

torkelo:
> I think this doesn't cover async queries (I discovered them recently and I'd say this is the only reason why we need this complicated code). An example of data source using them is the Athena DS.

@AgnesToulet All queries are async, what is special about the Athena DS?

AgnesToulet:
> @AgnesToulet All queries are async, what is special about the Athena DS?

When you use the TestData data source with the "Slow query" scenario, you can see the `query` request in a "pending" state for a while and returning after a specific time. This is caught by `waitForNetworkIdle` because it's waiting for all pending requests to end. 

With the Athena DS, a first `query` call is made, but it returns pretty quickly. Then, when the results are ready, another `query` call is made to retrieve them. If the time between these two requests is greater than the [`idleTime` option](https://pptr.dev/api/puppeteer.waitfornetworkidleoptions), `waitForNetworkIdle` will return before this second request is made.

torkelo:
@AgnesToulet is this some closed source athena data source? this ones does nothing special https://github.com/grafana/athena-datasource/blob/main/src/datasource.ts#L110C18-L110C23

torkelo:
@AgnesToulet oh, I see it's using this https://github.com/grafana/grafana-async-query-data-js/blob/main/src/DatasourceWithAsyncBackend.ts

src/browser/browser.ts
```diff
@@ -588,3 +556,55 @@ export class Browser {
...
+    __grafanaRunningQueryCount?: number;
+  }
+}
+
+async function waitForQueriesAndVisualizations(page: puppeteer.Page, options: ImageRenderOptions) {
+  await page.waitForFunction(
+    (isFullPage) => {
+      if (window.__grafanaSceneContext) {
+        return window.__grafanaRunningQueryCount === 0;
+      }
```
torkelo, line 572: @AgnesToulet pushed an update that restores the old waiting logic, except for scenes it will use this window query counter (That is not related to network query state but actual data source requests observable state and responses).  So for Athena this will work as they are returning observable loading states (And their observable does not complete until they they are done)

torkelo:
@AgnesToulet 
after https://github.com/grafana/grafana/pull/82441 is merged this should now work for old and scenes dashboards & panels

src/browser/browser.ts
torkelo: @AgnesToulet it works for every slow query I have tested for any data source, in both new and old dashboard arch. 

What version of Grafana are you using and what is the url? is the feature toggle __feature.dashboardSceneSolo enabled (or in url)?

AgnesToulet: @torkelo I tried to access the associated `/d-solo` page, and I'd say the issue is from there. The panel using the TestData DS with the slow query scenario always shows the loading page (the query is supposed to run for 15 seconds, but even after that time, I still see the loading page), and I don't see any query pending. 

So, I will approve the PR and will work on releasing it along with the PDF option next week.

AgnesToulet (approved):
LGTM!

src/browser/browser.ts
torkelo: @AgnesToulet strange, d-solo route works well for me no matter the panel or query

torkelo: @AgnesToulet you you using latest main?

AgnesToulet: @torkelo Actually, I tried again with a simple slow panel, and it's working. The panel that doesn't work is a repeating panel (using a custom variable).

torkelo: @torkelo for repeated panels you need use a new syntax for the panelId , example `panel-2-clone-2` , it's one of the few breaking changes in scenes

PR #522, opened 2024-05-06
CI: Fix publish to gcom step

AgnesToulet:
I tested it without the token to not publish wrong data and the command got executed here: https://drone.grafana.net/grafana/grafana-image-renderer/333/1/13 (but failed because of the token).


juanicabanas:
Does this already work? Do we still need it?

evictorero (approved):

Issue #505, opened 2024-03-08
Error Loading plugin in Gafana 10.4.0


fatbasstard:
Update our Grafana from `10.3.3` to `10.4.0` (not other changes) and now hitting the following error:

```
Error: ✗ *rendering.RenderingService run error: Unrecognized remote plugin message: 
Failed to read any lines from plugin's stdout
This usually means
the plugin was not compiled for this architecture,
the plugin is missing dynamic-link libraries necessary to run,
the plugin is not executable by this process due to file permissions, or
the plugin failed to negotiate the initial go-plugin protocol handshake
Additional notes about plugin:
Path: /var/lib/grafana-plugins/grafana-image-renderer/plugin_start_linux_amd64
Mode: -rwxr-xr-x
Owner: 472 [grafana] (current: 472 [grafana])
Group: 0 [root] (current: 0 [root])
ELF architecture: EM_X86_64 (current architecture: amd64)
```

Something changed in Grafana perhaps? 

See: https://github.com/grafana/grafana/blob/main/CHANGELOG.md#1040-2024-03-06

AgnesToulet:
Hello! I could not reproduce with Grafana 10.4.0 and the grafana-image-renderer 3.10.0 on Ubuntu. 

Which version of the grafana-image-renderer are you using? How do you run Grafana (on which OS? Using kubernetes? Docker? If so, could you share your configuration file?)

MujahidRashid:
I am also facing the same issue
Grafana version 10.3.3 running in portainer
Using this dockerfile to build the image
https://github.com/grafana/grafana/blob/main/packaging/docker/custom/Dockerfile

fatbasstard:
> Hello! I could not reproduce with Grafana 10.4.0 and the grafana-image-renderer 3.10.0 on Ubuntu.
> 
> Which version of the grafana-image-renderer are you using? How do you run Grafana (on which OS? Using kubernetes? Docker? If so, could you share your configuration file?)

Same as @MujahidRashid basically. We build our custom Docker image like this:

```docker
FROM grafana/grafana:10.4.0

RUN grafana cli --pluginsDir "$GF_PATHS_PLUGINS" --pluginUrl "https://github.com/grafana/grafana-image-renderer/releases/latest/download/plugin-linux-x64-glibc-no-chromium.zip" plugins install grafana-image-renderer
```

fatbasstard:
[...]

PR #525, opened 2024-05-21
Packages: Release Alpine package without Chromium

AgnesToulet:
Since Alpine released its latest version (3.19.x), the Linux plugin that runs with `glibc` has not been working anymore. To fix this, we should replace `glibc` with `gcompat`, but this could break support for other plugins. Hence, this seems better to release a new package.

This can't replace the Linux package because it doesn't work on Linux distributions that don't work with `musl` (tested on OpenSUSE and it's not working, I guess it would be the same on RHEL). 

Part of fixing https://github.com/grafana/grafana-image-renderer/issues/505 (Once this PR is merged, this should be updated to use it: https://github.com/grafana/grafana/blob/main/packaging/docker/custom/Dockerfile#L37)

Another solution could be to switch the Grafana image used in the [custom image](https://github.com/grafana/grafana/blob/main/packaging/docker/custom/Dockerfile) we are providing to `grafana/grafana:${GRAFANA_VERSION}-ubuntu` and stop supporting Alpine custom images.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -439,6 +439,7 @@ steps:
   - package-darwin-x64-unknown
   - package-win32-x64-unknown
   - package-linux-x64-glibc-no-chromium
+  - package-alpine-x64-no-chromium
   environment:
     GITHUB_TOKEN:
       from_secret: github_token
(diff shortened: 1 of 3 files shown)


DanCech (approved):
LGTM

PR #542, opened 2024-07-11
Full page image: Fix wait condition for dashboard with rows 

AgnesToulet:
This PR fixes the wait condition for dashboards with rows as it's been broken since [this change](https://github.com/grafana/grafana/pull/88768).

I know that relying on the `data-testid` is not great, but the test should alert us sooner next time (this should ideally also be covered by an E2E test in Grafana), and this condition is only used when dashboards are not rendered with Scenes (it will be removed when we stop supporting legacy dashboards).


evictorero (approved):
Looks good. If I understood correctly, this change make it work for Grafana 11.1 (when the CSS PR was merged) and older versions (since `data-testid` wasn't introduced in that PR).

AgnesToulet:
Ah good catch, the `data-testid` was added in 9.1 (in [this PR](https://github.com/grafana/grafana/pull/49722)) and the image renderer is still supporting [Grafana 8.3.11](https://github.com/grafana/grafana-image-renderer/blob/master/plugin.json#L31) so I updated my PR to support both (I don't think it's worth doing a breaking change).

PR #551, opened 2024-08-02
Chore: Don't install DevPackages in docker image 

McTonderski:
To remove possible security concern points I want to remove installing devPackages which are not needed in docker image.

Any suggestion is highly anticipated

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,7 @@ FROM base as build
 
 COPY . ./
 
-RUN yarn install --pure-lockfile
+RUN yarn install --pure-lockfile --production=true
 RUN yarn run build
 
 EXPOSE 8081
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,4 @@
+# Base stage
 FROM node:18-alpine AS base
 
 ENV CHROME_BIN="/usr/bin/chromium-browser"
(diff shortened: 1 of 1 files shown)


juanicabanas (approved):

Dockerfile
```diff
@@ -1,26 +1,34 @@
...
+FROM base AS dev-dependencies
 
+COPY package.json yarn.lock ./
 RUN yarn install --pure-lockfile
+
+# Build stage
+FROM dev-dependencies AS build
+
+COPY . ./
 RUN yarn run build
```
AgnesToulet, line 23:
```suggestion
# Build stage
FROM base AS build

COPY . ./

RUN yarn install --pure-lockfile
RUN yarn run build
```

AgnesToulet: @McTonderski Could you take a look at this suggestion, please? It simplifies a bit the code while still keeping the dev dependencies out of the final image

AgnesToulet:
Based on this PR, I opened and merged this one that contains a few updates: https://github.com/grafana/grafana-image-renderer/pull/575

Thanks for starting the work on this!

Dockerfile
McTonderski: hey, Thank you for pushing that MR !

PR #552, opened 2024-08-07
Scroll correct element when `bodyScrolling` is enabled

ashharrison90:
On non-scenes dashboards when `bodyScrolling` is enabled, we need to scroll the window.

Fixes https://github.com/grafana/grafana-enterprise/issues/6938

### DIFF
--- a/src/browser/browser.ts
+++ b/src/browser/browser.ts
@@ -578,11 +578,6 @@ declare global {
   interface Window {
     __grafanaSceneContext: object;
     __grafanaRunningQueryCount?: number;
-    grafanaBootData: {
-      settings: {
-        featureToggles: Record<string, boolean | undefined>
-      };
-    };
   }
 }
 
--- a/src/browser/browser.ts
+++ b/src/browser/browser.ts
@@ -176,25 +176,29 @@ export class Browser {
   }
 
   async scrollToLoadAllPanels(page: puppeteer.Page, options: ImageRenderOptions): Promise<DashboardScrollingResult> {
-    const scrollDivSelector = '#page-scrollbar,[class*="scrollbar-view"]';
+    const scrollElementSelector = await page.evaluate(() => {
+      const pageScrollbarSelector = '#page-scrollbar';
+      const hasPageScrollbar = Boolean(document.querySelector(pageScrollbarSelector));
+      return hasPageScrollbar ? pageScrollbarSelector : 'body';
+    });
     const scrollDelay = options.scrollDelay ?? 500;
 
-    await page.waitForSelector(scrollDivSelector);
-    const heights: { dashboard?: { scroll: number; client: number }; body: { client: number } } = await page.evaluate((scrollDivSelector) => {
+    await page.waitForSelector(scrollElementSelector);
+    const heights: { dashboard?: { scroll: number; client: number }; body: { client: number } } = await page.evaluate((scrollElementSelector) => {
       const body = { client: document.body.clientHeight };
-      const scrollableDiv = document.querySelector(scrollDivSelector);
-      if (!scrollableDiv) {
-        this.log.debug('no scrollable div detected, returning without scrolling')
+      const scrollableElement = document.querySelector(scrollElementSelector);
+      if (!scrollableElement) {
+        this.log.debug('no scrollable element detected, returning without scrolling')
         return {
           body,
         };
       }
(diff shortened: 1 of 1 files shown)


src/browser/browser.ts
```diff
@@ -176,25 +176,29 @@ export class Browser {
   }
 
   async scrollToLoadAllPanels(page: puppeteer.Page, options: ImageRenderOptions): Promise<DashboardScrollingResult> {
-    const scrollDivSelector = '#page-scrollbar,[class*="scrollbar-view"]';
+    const scrollElementSelector = await page.evaluate(() => {
+      const pageScrollbarSelector = '#page-scrollbar';
```
evictorero, line 180: Why the `[class*="scrollbar-view"]` was removed? Will this change support older versions of Grafana like 9 and 10?

src/browser/browser.ts
```diff
@@ -213,19 +217,23 @@ export class Browser {
...
-        (scrollByHeight, scrollDivSelector) => {
-          document.querySelector(scrollDivSelector)?.scrollBy(0, scrollByHeight);
+        (scrollByHeight, scrollElementSelector) => {
+          scrollElementSelector === 'body'
+            ? window.scrollBy(0, scrollByHeight)
+            : document.querySelector(scrollElementSelector)?.scrollBy(0, scrollByHeight);
         },
         heights.dashboard.client,
-        scrollDivSelector
+        scrollElementSelector,
```
evictorero, line 226: This code now has more logic, and it is the same as below. Can we extract this and call it with height 0 in the latter case?

ashharrison90: one uses `scrollBy` and the other uses `scrollTo`

i guess we could change the first use to `scrollTo` and multiply the loop index, but i was trying to avoid modifying the existing code tbh

src/browser/browser.ts
ashharrison90: yeah so this is tricky. we need a way to find if the **page** has a scrollbar container. i'm trying to avoid using the feature toggle, because when we then enable that feature toggle by default and remove it we'll run into a problem where we need to update this code as well. but doing `[class*="scrollbar-view"]` is just so generic it won't work. it picks up scrollbar containers in panels, graph legends, etc. the `#page-scrollbar` id was added in grafana 11.1... i can make the `scrollbar-view` selector more specific (e.g. something like `#page-content > div > div > [class="scrollbar-view"]`), but then we'd have to add another different selector for scenes dashboards since it's slightly different and as you can see these selectors would be pretty brittle 😬 any suggestions? 🤔

evictorero: It is tricky 😞. I remember we made a change [here](https://github.com/grafana/grafana-image-renderer/pull/524) after Grafana 11 (probably because of the new #page-scrollbar) and just want to confirm this will not introduce a regression in an older Grafana version.
We need to support retro compatibility with the latest 3 majors and all feature toggles, with and without Scenes.

ashharrison90: @evictorero ok i've added some horrible code to add multiple more specific fallback selectors

i'm pretty sure this will now be fully backwards/forwards compatible, it's just... horrible 😂 but i can't really think of a better way of doing it.

juanicabanas:
LGTM. Something to take into account:
This `scrollToLoadAllPanels` function is executed to load all panels before taking the screenshot. After some discussion due to [this escalation](https://github.com/grafana/support-escalations/issues/11432), I think we should not lazy load the panels when we need a screenshot, because we know we want all panels to load. Therefore, this code should be deleted in a near future

juanicabanas (approved):

ashharrison90:
@juanicabanas @evictorero what's the process like for releasing a new version of the plugin? is there anything i should do in this PR? 🤔

evictorero:
> @juanicabanas @evictorero what's the process like for releasing a new version of the plugin? is there anything i should do in this PR? 🤔

You need to create a PR with a few changes, including the version and the changelog, and then the CI/CD creates the release after merging it. You can check the last commit for inspiration or we can do it for you.

ashharrison90:
> You need to create a PR with a few changes, including the version and the changelog, and then the CI/CD creates the release after merging it. You can check the last commit for inspiration or we can do it for you.

that's ok, i can do it. thanks for the pointers 👍

PR #554, opened 2024-08-13
Full page image: Fix blank page 

juanicabanas:
This PR fixes blank page screenshot bug.

Puppeter resizes the viewport when taking a screenshot. See details [here](https://stackoverflow.com/questions/68059664/puppeteer-page-screenshot-resizes-viewport)

After [this PR](https://github.com/grafana/grafana/pull/90755) in scenes, when taking the screenshot `isMobile` is true for a while (randomly). That makes the dashboard re-render and the screenshot ends up being a blank page, showing only the NavToolbar.
[Issue thread](https://raintank-corp.slack.com/archives/C03KVDHTWAH/p1722456214228329).

With `captureBeyondViewport` as false, it works as expected, because we set the viewport based on [panels container `scrollHeight`](https://github.com/grafana/grafana-image-renderer/blob/master/src/browser/browser.ts#L211) [here](https://github.com/grafana/grafana-image-renderer/blob/master/src/browser/browser.ts#L371)

| Before  | After |
| ------------- | ------------- |
| image | image |

### DIFF
--- a/src/browser/browser.ts
+++ b/src/browser/browser.ts
@@ -130,6 +130,7 @@ export class Browser {
       ignoreHTTPSErrors: this.config.ignoresHttpsErrors,
       dumpio: this.config.dumpio,
       args: this.config.args,
+      defaultViewport: null
     };
 
     if (this.config.chromeBin) {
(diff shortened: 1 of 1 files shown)


torkelo (approved):

evictorero (approved):
Tested and couldn't reproduce the blank page anymore. It takes good screenshots with and without scenes.

juanicabanas:
https://github.com/grafana/grafana-enterprise/issues/7056

lsfc02:
the same thing here

juanicabanas:
Hi @lsfc02 . Are you still facing the same issue?
Could you provide more info? 
Thanks :)

Issue #460, opened 2023-09-19
Browser: Switch to new headless mode


AgnesToulet:
Chrome is releasing a [new headless](https://developer.chrome.com/articles/new-headless/) mode. It is currently known to have some issues including [one](https://github.com/puppeteer/puppeteer/issues/10829) that seems to make Chrome crash when using Kubernetes so we are [not using it](https://github.com/grafana/grafana-image-renderer/pull/459).

In the future though, the old headless mode will be removed and we will need to either switch to the new headless mode or to use different Chrome binaries that contain the old one.

evictorero:
There was a change in the [strategy](https://developer.chrome.com/blog/chrome-headless-shell), we could continue using the old headless mode for performance.
I'm planning to add it as part of the upgrade to Puppeteer 22.

PR #556, opened 2024-08-15
Puppeteer: Upgrade to v22

evictorero:
* Upgrade Puppeteer to the next [major](https://github.com/puppeteer/puppeteer/blob/main/packages/puppeteer/CHANGELOG.md#2200-2024-02-05) with support.
* Bump axios

We are doing this to keep using a maintained version and remove some old, unmaintained dependencies.

Fix https://github.com/grafana/grafana-image-renderer/issues/460


src/browser/reusable.ts
```diff
@@ -23,7 +23,7 @@ export class ReusableBrowser extends Browser {
     try {
       page = await this.withTimingMetrics<puppeteer.Page>(async () => {
         this.validateImageOptions(options);
-        context = await this.browser.createIncognitoBrowserContext();
+        context = await this.browser.createBrowserContext();
```
evictorero, line 26: This function was renamed on version 22

package.json
```diff
@@ -28,7 +28,6 @@
     "@grpc/grpc-js": "^1.8.22",
     "@grpc/proto-loader": "^0.7.2",
     "@hapi/boom": "^10.0.0",
-    "@puppeteer/browsers": "^1.6.0",
```
evictorero, line 31: I removed this to use the dependency included inside Puppeteer core. I couldn't find a reason why we were pinning this dependency here.

juanicabanas (approved):

AgnesToulet:
This new Puppeteer version also enables the [new-headless mode by default](https://github.com/puppeteer/puppeteer/pull/11815) that [doesn't work for Kubernetes users](https://github.com/grafana/grafana-image-renderer/issues/460). If we want to merge this quickly, we should use `headless=chrome-headless-shell` [here](https://github.com/grafana/grafana-image-renderer/blob/master/src/browser/browser.ts#L140), else I'll spend some time investigating the impact of this.

Also, I noticed an issue with the stat panel in the test screenshots, I'm looking into this.

src/app.ts
```diff
@@ -16,9 +16,6 @@ async function main() {
   const env = Object.assign({}, process.env);
   const command = argv._[0];
 
-  // See https://github.com/grafana/grafana-image-renderer/issues/460
```
evictorero, line 19: This is no longer needed. We are using the `shell` value instead of the new headless mode.

src/browser/browser.ts
```diff
@@ -137,7 +137,7 @@ export class Browser {
       launcherOptions.executablePath = this.config.chromeBin;
     }
 
-    launcherOptions.headless = !this.config.headed;
+    launcherOptions.headless = !this.config.headed ? "shell" : false;
```
evictorero, line 140: shell is for the old headed [mode](https://developer.chrome.com/blog/chrome-headless-shell)

src/app.ts
AgnesToulet: Good catch! This is also set in the tests: https://github.com/grafana/grafana-image-renderer/blob/master/src/service/http-server.integration.test.ts#L101

src/browser/browser.ts
AgnesToulet, line 140: Nit
```suggestion
    launcherOptions.headless = !this.config.headed ? 'shell' : false;
```

AgnesToulet (approved):
LGTM! This looks good and works fine 👍 

There is a bug in the tests, it's why the screenshot with the error wasn't caught. I'll work on a separate PR to fix this. 

Note: There is also [this change](https://github.com/puppeteer/puppeteer/pull/11778) which updates a setting in PDF. We may want to make this configurable as this can increase the file size but this would also be in another PR.

PR #583, opened 2024-11-28
Chore: Add ENV variables for temp folders

evictorero:
This PR defines 2 ENV variables to let Puppeteer write temporal files in Chromium.
A change introduced in a Chromium version released with the image renderer 3.11.1 broke the image renderer starting process in some Linux environments.
This problem is not present in the newest versions 3.11.5 and 3.11.6. We are adding this to the Dockerfile to prevent future errors in case chromium gets updated again.

Fixes https://github.com/grafana/grafana-image-renderer/issues/545
Fixes https://github.com/grafana/grafana/pull/97067


nmarrs (approved):
LGTM

PR #586, opened 2024-12-10
Image Render: Support Tracing

lucychen-grafana:
Testing Instructions:

**Test as service:**

1. Install tempo https://grafana.com/docs/tempo/latest/getting-started/docker-example/
- Run the dockerfile for[ tempo local storage](https://github.com/grafana/tempo/blob/main/example/docker-compose/local/readme.md) in the local directory

2. Checkout [this branch](https://github.com/grafana/grafana/pull/100559) of grafana and run grafana-enterprise
- Set otlp address in defaults.ini 
```
[tracing.opentelemetry.otlp]
address = localhost:4317
```
3. Checkout this pr grafana-image-renderer branch. Run as remote server 

4. Open Localhost:3000; run a report or PDF creation, or any call to the image renderer service

5. Check traces: Explore -> Select ‘gdev-tempo’ datasource -> run query type Search with `resource.service.name="grafana-image-renderer"`
- If there is no Tempo datasource, create a new one with url set to `http://localhost:3200`
- Select the TraceID related to the IR GET request and look into the spans

6. Check logs in the console to confirm trace_id and span_id is included in the logs

**To test the plugin on Mac:**
1. Build the plugin locally for Mac:
    1. Update your `Makefile` to set `ARCH = darwin-arm64-unknown` 
    1. Run `make build_package`
    1. The plugin will be built in the `dist` folder
2. Copy / paste the built plugin to your Grafana plugins folder and rename it to `grafana-image-renderer`
3. In your Grafana .ini file:
    1. Comment your remote renderer settings in the `[rendering]` section
    2. Add the following config to enable traces and allow running the IR plugin unsigned:
```
[plugin.grafana-image-renderer]
rendering_tracing_url = http://localhost:4318/v1/traces

[plugins]
allow_loading_unsigned_plugins = grafana-image-renderer
```

Additional Pr to review:
[Deployment_tools pr ](https://github.com/grafana/deployment_tools/pull/216067) for config setup
[Grafana pr](https://github.com/grafana/grafana/pull/100559)


src/logger.ts
```diff
@@ -45,6 +46,8 @@ export class ConsoleLogger implements Logger {
       transports.push(new winston.transports.Console(options));
     }
 
+//@opentelemetry/instrumentation-winston auto inject trace-context into Winston log records 
+
```
lucychen-grafana, line 50: Confirm this is the case in the log output in terminal. You can check that `trace_id`, `span_id`, and `trace_flags` is included

src/tracing.ts
```diff
@@ -0,0 +1,41 @@
+import {NodeSDK} from '@opentelemetry/sdk-node';
```
lucychen-grafana, line 1: I followed [OpenTelemetry Doc](https://opentelemetry.io/docs/languages/js/getting-started/nodejs/) for Node.js to implement this

src/tracing.ts
```diff
@@ -0,0 +1,41 @@
...
+
+const sdk = new NodeSDK({
+    resource: new Resource({
+        [SEMRESATTRS_SERVICE_NAME]: 'image-renderer-service',
+    }),
+    traceExporter,
+    instrumentations: [
+        getNodeAutoInstrumentations({
+        // only instrument fs if it is part of another trace
+        '@opentelemetry/instrumentation-fs': {
```
lucychen-grafana, line 24: the [fs module](https://nodejs.org/dist/v23.6.1/docs/api/fs.html) is very noisy. It create 10+ traces with singular span so it did not make much sense to have it unless it's part of another trace.

src/tracing.ts
```diff
@@ -0,0 +1,41 @@
...
+export  function startTracing(log: Logger) {
+    sdk.start();
+    log.info('Starting tracing');
+
+    process.on('SIGTERM', () => {
+        sdk.shutdown()
+            .then(() => log.debug('Tracing terminated'))
+            .catch((error) => log.error('Error terminating tracing', error))
+            .finally(() => process.exit(0));
+    });
```
lucychen-grafana, line 40: Create startTracing separately from the rest of initialization steps b/c we want to only capture traces if it's[ enabled](https://github.com/grafana/grafana-image-renderer/pull/586/files#diff-841254fe75488c1bd4cd7f68f00b4be0e48dcfbc4a16b45847b68295e0e3b27bR69-R72)

src/app.ts
```diff
@@ -1,3 +1,4 @@
+import { startTracing } from './tracing';
```
lucychen-grafana, line 1: OpenTelemetry needs to be initialized before the other modules (express, http, winston, etc) in order for it to auto capture telemetry trace data from those libraries so it must come first

lucychen-grafana:
> [tracing.opentelemetry.otlp]
> # otlp destination (ex localhost:4317)
> address = localhost:4317
> ```

@AgnesToulet setting this address is enable traces in Grafana in dev env right? Traces should already be enabled in Cloud via Faro

AgnesToulet:
Yes it's for your local environment. It's enabled by default on Cloud (Faro is for frontend monitoring, its configuration is in a different section of the ini file, I don't think both are related).

README.md
```diff
@@ -40,6 +39,33 @@ This plugin is not compatible with the current Grafana Docker image and requires
 
 If you still want to install the plugin with the Grafana Docker image, refer to the instructions on building a custom Grafana image in [Grafana Docker documentation](https://grafana.com/docs/grafana/latest/setup-grafana/configure-docker/#build-a-custom-grafana-docker-image).
 
+### Plugin for Mac ARM64
```
AgnesToulet, line 42: I think this should be here instead: https://github.com/grafana/grafana-image-renderer/blob/master/docs/building_from_source.md

This README file is the one that will be published on [grafana.com](https://grafana.com/grafana/plugins/grafana-image-renderer/?tab=overview) and should stay simple I think (I just noticed that the "testing" section should also probably be elsewhere).

We could link to the "Build from source" doc in this file though, to help with discoverability. Like, in the "Supported operating systems" section, you could add:
```
For Mac ARM64, you need to [build the plugin from source](https://github.com/grafana/grafana-image-renderer/blob/master/docs/building_from_source.md) or use the [remote rendering installation](https://github.com/grafana/grafana-image-renderer?tab=readme-ov-file#remote-rendering-service-installation).
```

README.md
```diff
@@ -73,6 +99,11 @@ The following example shows how you can run Grafana and the remote HTTP renderin
        ports:
          - 8081
    ```
+   _Notes:_
```
AgnesToulet, line 102: This command is already in the "build from source" section.

devenv/docker/custom-config/config.json
```diff
@@ -41,6 +41,10 @@
     },
 
     "verboseLogging": false,
-    "dumpio": false
+    "dumpio": false,
+
+    "tracing": {
+      "url": "http://host.docker.internal:4318/v1/traces"
```
AgnesToulet, line 47: This works only on Windows and Mac, not Linux. I think the best would be to provide the whole setup so to copy/paste the content of the Tempo docker-compose file, to replace this with `http://tempo:4318/v1/traces` and to provision the Tempo datasource so people can easily set up tracing (I wonder if we should do all this in the `custom-config` folder or create a new `tracing` folder).

src/plugin/v2/grpc_plugin.ts
```diff
@@ -251,6 +237,27 @@ class PluginGRPCServer {
...
+
+  getHeaders(req: RenderRequest | RenderCSVRequest): HTTPHeaders {
+    const headers: HTTPHeaders = {};
+
+    if (req.headers) {
+      if (req.headers.hasOwnProperty('Accept-Language')) {
+        const h = req.headers['Accept-Language'];
+        headers['Accept-Language'] = h.values.join(';');
+      }
+    }
```
AgnesToulet, line 249:
```suggestion
    if (req.headers?.hasOwnProperty('Accept-Language')) {
      const h = req.headers['Accept-Language'];
      headers['Accept-Language'] = h.values.join(';');
    }
```

src/tracing.ts
```diff
@@ -0,0 +1,52 @@
...
+  // const { diag, DiagConsoleLogger, DiagLogLevel } = require('@opentelemetry/api');
+  // diag.setLogger(new DiagConsoleLogger(), DiagLogLevel.DEBUG);
+
+  const traceExporter = new OTLPTraceExporter({
+    url: exporterURL,
+  });
+
+  return new NodeSDK({
+    resource: new Resource({
+      [SEMRESATTRS_SERVICE_NAME]: 'grafana-image-renderer',
```
AgnesToulet, line 27: As mentioned in a previous comment, I think it would be best to be able to configure this.

README.md
lucychen-grafana: Ok. I can move the instruction to Build from source doc and also create a new doc for testing instructions

devenv/docker/custom-config/config.json
lucychen-grafana: Oh i see. Maybe better to create separate tracing folder w tracing config.json and docker-compose file if end user don't need tracing enabled at all times

src/tracing.ts
lucychen-grafana: Configure the [SEMRESATTRS_SERVICE_NAME]? Or the whole resource?

AgnesToulet: The [SEMRESATTRS_SERVICE_NAME], I think internally, we'd prefer to use `render-service` for example. But `grafana-image-renderer` is a better default.

lucychen-grafana: Why do we want to give that an option to configure it? If in the future we decide to rename it, then it'll make querying disjuncted and would have to take into account of both

AgnesToulet: It's not to configure it multiple times, it's just to configure it once when you set up tracing so that it matches your own deployment convention and make the traces easier to discover. The thing is that the value you put makes the most sense as the default but doesn't match what we use on our Cloud.

devenv/docker/tracing/docker-compose.yaml
```diff
@@ -0,0 +1,74 @@
...
+    environment:
+      GF_RENDERING_SERVER_URL: http://renderer:8081/render
+      GF_RENDERING_CALLBACK_URL: http://grafana:3000/
+      GF_RENDERING_TRACING_URL: http://tempo:4318/v1/traces
+      GF_LOG_FILTERS: rendering:debug
+      GF_AUTH_ANONYMOUS_ENABLED: true
+      GF_AUTH_ANONYMOUS_ORG_ROLE: Admin
+      GF_AUTH_DISABLE_LOGIN_FORM: true
+      GF_FEATURE_TOGGLES_ENABLE: traceqlEditor metricsSummary
+      GF_INSTALL_PLUGINS: https://storage.googleapis.com/integration-artifacts/grafana-exploretraces-app/grafana-exploretraces-app-latest.zip;grafana-traces-app
```
AgnesToulet, line 66:
```suggestion
      GF_INSTALL_PLUGINS: https://storage.googleapis.com/integration-artifacts/grafana-exploretraces-app/grafana-exploretraces-app-latest.zip;grafana-traces-app
      GF_TRACING_OPENTELEMETRY_OTLP_ADDRESS: tempo:4317
```
This is hard to test because the work on Grafana isn't merged yet, but I'd say this is necessary to get Grafana traces and see the full traces from / to Grafana.

src/config/rendering.ts
```diff
@@ -13,6 +13,11 @@ type NetworkConditions = {
   latency: number;
 };
 
+export interface TracesConfig {
+  url: string;
+  serviceName: string;
```
AgnesToulet, line 18:
```suggestion
  serviceName?: string;
```
I'd put it as optional to not have to fill them everywhere, especially in the envConfig object.

src/config/rendering.ts
```diff
@@ -118,6 +132,11 @@ const envConfig: Record<Mode, Keys<RenderingConfig>> = {
     verboseLogging: 'GF_PLUGIN_RENDERING_VERBOSE_LOGGING',
     dumpio: 'GF_PLUGIN_RENDERING_DUMPIO',
     timingMetrics: 'GF_PLUGIN_RENDERING_TIMING_METRICS',
+    tracing: {
+      url: 'GF_PLUGIN_RENDERING_TRACING_URL',
+      serviceName: 'GF_PLUGIN_RENDERING_TRACING_SERVICE_NAME'
```
AgnesToulet, line 137: This is not used by the `populateRenderingConfigFromEnv` function. This should either be used or removed. I think we can remove it as it's less likely to be an issue for people using the image renderer in plugin mode (usually, this means no Kubernetes setup where you already gave a name at your services). And we can always add it later if requested.

src/service/http-server.integration.test.ts
```diff
@@ -95,15 +99,15 @@ function getGrafanaEndpoint(domain: string) {
 let envSettings = {
   saveDiff: false,
   updateGolden: false,
-}
+};
```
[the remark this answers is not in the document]
lucychen-grafana: yep I can create new issue for that

src/service/http-server.ts
```diff
@@ -330,4 +318,20 @@ export class HttpServer {
       return res.status(500).json({ error: e.message });
     }
   };
+
+  getHeaders(req: express.Request<any, any, any, RenderOptions, any>): HTTPHeaders {
```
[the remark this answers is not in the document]
lucychen-grafana: There express module Request identifies some params as any:
```
    interface Request<
        P = core.ParamsDictionary,
        ResBody = any,
        ReqBody = any,
        ReqQuery = core.Query,
        Locals extends Record<string, any> = Record<string, any>,
    > extends core.Request<P, ResBody, ReqBody, ReqQuery, Locals> {}
```

lucychen-grafana:
> One thing I was a little confused on is whether or not traces should work for image renderer in plugin mode or just http server mode (I think the latter). Your directions in PR description include instructions for mac for running as plugin mode but I'm not sure traces will be sent in that case 🤔

In Cloud it is running in server mode. Originally was going just do that but Agnes was able to add plugin mode since the change wasn't hard.

Issue #186, opened 2020-11-02
Server doesn't cancel browser rendering when client cancel


marefr:
**What happened**:
Error message logged in Grafana `t=2020-10-23T14:56:24+0000 lvl=eror msg="Failed to get render key from cache" logger=rendering renderer=http error="cache item not found"`

**What you expected to happen**:
No error message.

**How to reproduce it (as minimally and precisely as possible)**:
Render a panel and refresh the request in browser multiple times frequently.

**Anything else we need to know?**:
After investigation it has been concluded that browser rendering is not cancelled server side when client (Grafana) cancel a request.

**Environment**:
- Grafana Image Renderer version: master / 2.0.0
- Grafana version: master
- Installed plugin or remote renderer service: Both

PR #588, opened 2024-12-17
Support cancel rendering on client cancellation

AgnesToulet:
Currently when client cancels the requests, the rendering request continues to run until it ends. 

This PR fixes that by stopping the rendering process.

It has two main benefits:
- Improve performance when the image renderer is receiving a lot of requests from Grafana, especially when it comes from the alerting feature. In the alerting feature, image renderer requests are stacked when they exceed the concurrent request limit, so they can be made after a few seconds, be cancelled right after that because they exceed the timeout and still use the image renderer CPU and memory to render the image.
- Remove logs `Failed to get render key from cache`. This is displayed when the image renderer tries to access Grafana after the request is cancelled because the render key has been deleted but this makes it more complicated to investigate real issues.

Fixes https://github.com/grafana/grafana-image-renderer/issues/186

### DIFF
--- a/src/browser/browser.ts
+++ b/src/browser/browser.ts
@@ -9,7 +9,7 @@ import * as Jimp from 'jimp';
 import { Logger } from '../logger';
 import { RenderingConfig } from '../config/rendering';
 import { ImageRenderOptions, RenderOptions } from '../types';
-import { TimeoutError } from './error';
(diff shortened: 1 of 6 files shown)


juanicabanas (approved):
Great work! Tested and works as expected 🚀

PR #589, opened 2024-12-24
Logs: Redirect log in response

juanicabanas:


src/browser/browser.ts
```diff
@@ -561,6 +563,14 @@ export class Browser {
     this.log.debug('Browser request', 'url', req.url(), 'method', req.method());
   };
 
+  logRedirectResponse = (resp: puppeteer.HTTPResponse) => {
+    const status = resp.status()
+    if (status >= 300 && status <= 399) {
+      const headers = resp.headers();
+      this.log.debug(`Redirect from ${resp.url()} to ${headers['location']}`)
```
AgnesToulet, line 570: It seems like one redirect request contains this `Location` header while the others do not, leading to a few `Redirect from xxx to undefined` logs.
```suggestion
      const redirectLocation = headers['location'];
      if (redirectLocation) {
        this.log.debug(`Redirect from ${resp.url()} to ${redirectLocation}`);
      }
```

src/browser/browser.ts
```diff
@@ -561,6 +563,14 @@ export class Browser {
     this.log.debug('Browser request', 'url', req.url(), 'method', req.method());
   };
 
+  logRedirectResponse = (resp: puppeteer.HTTPResponse) => {
+    const status = resp.status()
```
AgnesToulet, line 567:
```suggestion
    const status = resp.status();
```

AgnesToulet (approved):
LGTM! Tested it and it works fine but it generates a few logs that contains `undefined` instead of the redirect URL so I left a suggestion to fix this.

src/browser/browser.ts
juanicabanas: why are there redirects without Location header? Shouldn't those be also logged to understand the flow?

src/browser/browser.ts
```diff
@@ -518,6 +518,7 @@ export class Browser {
       page.on('request', this.logRequest);
       page.on('requestfinished', this.logRequestFinished);
       page.on('close', this.logPageClosed);
+      page.on('response', this.logRedirectResponse)
```
AgnesToulet, line 521:
```suggestion
      page.on('response', this.logRedirectResponse);
```

src/browser/browser.ts
```diff
@@ -531,6 +532,7 @@ export class Browser {
       page.off('request', this.logRequest);
       page.off('requestfinished', this.logRequestFinished);
       page.off('close', this.logPageClosed);
+      page.off('response', this.logRedirectResponse)
```
AgnesToulet, line 535:
```suggestion
      page.off('response', this.logRedirectResponse);
```

src/browser/browser.ts
```diff
@@ -561,6 +563,14 @@ export class Browser {
     this.log.debug('Browser request', 'url', req.url(), 'method', req.method());
   };
 
+  logRedirectResponse = (resp: puppeteer.HTTPResponse) => {
+    const status = resp.status();
+    if (status >= 300 && status <= 399 && resp.request().resourceType() === 'document') {
+      const headers = resp.headers();
+      this.log.debug(`Redirect from ${resp.url()} to ${headers['location']}`)
```
AgnesToulet, line 570:
```suggestion
      this.log.debug(`Redirect from ${resp.url()} to ${headers['location']}`);
```

PR #602, opened 2025-02-18
Add tracing to plugin mode and improve config management

AgnesToulet:
This PR adds tracing to the plugin mode and update a few things from the original PR:
- Read the config file before initializing tracing, this allows to have the exporter URL set in the config file instead of using an env variable
- Do no intialize the `tracer` variable if tracing is disabled
- Propagate the trace context back to Grafana

**What remains:**
- [x] Add tracing to the CSV endpoint
- [x] Compare traces between the original PR and this one to ensure we didn't lose anything by reading the config file before initializing tracing
- [x] Update [Grafana upstream PR](https://github.com/grafana/grafana/pull/100559) 
- [ ] Test, test and test again (in plugin mode, in remote server mode and in Docker)

**To test the plugin on Mac:**
1. Build the plugin locally for Mac:
    1. Update your `Makefile` to set `ARCH = darwin-arm64-unknown` 
    1. Run `make build_package`
    1. The plugin will be built in the `dist` folder
2. Copy / paste the built plugin to your Grafana plugins folder and rename it to `grafana-image-renderer`
3. In your Grafana .ini file:
    1. Comment your remote renderer settings in the `[rendering]` section
    2. Add the following config to enable traces and allow running the IR plugin unsigned:
```
[plugin.grafana-image-renderer]
rendering_tracing_url = http://localhost:4318/v1/traces

[plugins]
allow_loading_unsigned_plugins = grafana-image-renderer
```

### DIFF
--- a/dev.json
+++ b/dev.json
@@ -15,10 +15,6 @@
         "json": false,
         "colorize": true
       }
-    },
-    
-    "tracing": {
-      "enabled": true
     }
   },
   "rendering": {
@@ -54,6 +50,11 @@
 
     "verboseLogging": true,
     "dumpio": false,
-    "timingMetrics": true
+    "timingMetrics": true,
+
+    "tracing": {
+      "enabled": true,
+      "url": "http://localhost:4318/v1/traces"
+    }
   }
 }
--- a/dev.json
+++ b/dev.json
@@ -53,8 +53,7 @@
     "timingMetrics": true,
 
     "tracing": {
-      "enabled": true,
-      "url": "http://localhost:4318/v1/traces"
+      "url": ""
     }
   }
 }
--- a/src/config/rendering.ts
+++ b/src/config/rendering.ts
@@ -14,7 +14,6 @@ type NetworkConditions = {
 };
 
 export interface TracesConfig {
-  enabled: boolean;
   url: string;
 }
 
@@ -68,7 +67,6 @@ export const defaultRenderingConfig: RenderingConfig = {
   dumpio: false,
   timingMetrics: false,
   tracing: {
-    enabled: false,
     url: '',
   },
 };
@@ -129,7 +127,6 @@ const envConfig: Record<Mode, Keys<RenderingConfig>> = {
     dumpio: 'GF_PLUGIN_RENDERING_DUMPIO',
     timingMetrics: 'GF_PLUGIN_RENDERING_TIMING_METRICS',
     tracing: {
-      enabled: 'GF_PLUGIN_RENDERING_TRACING_ENABLED',
       url: 'GF_PLUGIN_RENDERING_TRACING_URL',
     },
   },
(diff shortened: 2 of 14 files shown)


src/plugin/v2/grpc_plugin.ts
```diff
@@ -251,6 +237,27 @@ class PluginGRPCServer {
...
+    if (req.headers) {
+      if (req.headers.hasOwnProperty('Accept-Language')) {
+        const h = req.headers['Accept-Language'];
+        headers['Accept-Language'] = h.values.join(';');
+      }
+    }
+
+    if (this.config.rendering.tracing.url) {
+      const output: TraceCarrier = {};
+      propagation.inject(context.active(), output);
```
lucychen-grafana, line 253: You are referencing this [manual propagation](https://opentelemetry.io/docs/languages/js/propagation/#generic-example)correct?

src/config/rendering.ts
```diff
@@ -118,6 +126,9 @@ const envConfig: Record<Mode, Keys<RenderingConfig>> = {
     verboseLogging: 'GF_PLUGIN_RENDERING_VERBOSE_LOGGING',
     dumpio: 'GF_PLUGIN_RENDERING_DUMPIO',
     timingMetrics: 'GF_PLUGIN_RENDERING_TIMING_METRICS',
+    tracing: {
+      url: 'GF_PLUGIN_RENDERING_TRACING_URL',
```
lucychen-grafana, line 130: How is this linked to 
```
[plugin.grafana-image-renderer]
rendering_tracing_url
```
? Where is it being set?

lucychen-grafana:
I tested as a remote server and as a plugin and they look good. 
I tried running it on Docker but my setup might not be correct...the logs are not showing traces the trace url is set in the custom-config config.json

dev.json
```diff
@@ -54,6 +50,10 @@
 
     "verboseLogging": true,
     "dumpio": false,
-    "timingMetrics": true
+    "timingMetrics": true,
+
+    "tracing": {
+      "url": ""
```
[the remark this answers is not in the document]
AgnesToulet: We could as this is not generating errors when you don't have Tempo set up. We should update the doc though to explain how to set up the environment to get traces. We could add this URL there as well.

src/config/rendering.ts
AgnesToulet: In the Grafana codebase, the config fields for plugins (so in sections starting with `plugin.`) are automatically converted to env variables to be used by plugins. They follow the syntax `GF_PLUGIN_<config field in upper case>`

AgnesToulet:
To make it work in Docker, you need to update the URL to `http://host.docker.internal:4318/v1/traces`. This allows Docker services to communicate if they are not set up in the same `docker-compose` file. Maybe we could update the `custom-config` setup to include traces (and setup the tempo services in the same docker-compose file instead of linking to the Tempo repo).

PR #612, opened 2025-03-12
Image Render: Support tracing

lucychen-grafana:
Second attempt to add tracing. Updates from the original PR:
- Do not add tracing headers when tracing is disabled
- Do not add tracing headers for requests that go to a different hostname than the original request: that is not the most elegant solution but there is no easy way in Puppeteer to add headers only for specific requests (see https://github.com/puppeteer/puppeteer/issues/2627). Maybe we could remove it in the feature if we fix the issue we had on Cloud and if that's not impacting other users or make it configurable but I'd like to start with this to unblock us.

To see only the changes from the original PR: https://github.com/grafana/grafana-image-renderer/compare/tmp-tracing-base...lucychen/add_tracing_fix?expand=1


src/browser/browser.ts
```diff
@@ -575,6 +604,22 @@ export class Browser {
     }
   }
 
+  removeTracingHeader = (req: puppeteer.HTTPRequest) => {
```
evictorero, line 607: Wouldn't be safe to try/catch this code? I'm mostly concerned about `new URL(referer)` if it is not a URL it will fail but in general, it sounds safer to try/catch the entire block and take a decision if it fails (maybe just continue without deleting the headers

PR #618, opened 2025-03-27
The plugin could have the feature to create PNG-8 greyscaled pictures

devbar:
This feature is very usefull if grafana should render on an eInk display with greyscale only. I use it for a jailbreaked kindle.

This implementation does not allow to pass the _greyScaleImage_ flag in the render url. Maybe some of you can point me to the right line of code. Until this it's necessary to send the arg in the URL of the renderer in docker config.

```
 grafana:
    [...]
    environment:
      - GF_INSTALL_PLUGINS=grafana-image-renderer
      - GF_RENDERING_SERVER_URL=http://renderer:8081/render?greyScaleImage=true
    [...]
```


Proximyst:
Could you please sign the CLA? This would be required to get a proper review going

PR #595, opened 2025-02-13
Chore: Upgrade to Node 22

AgnesToulet:
This PR upgrades to Node 22 as Node 18 will reach end of life in April. 

To build the plugin using Node 22, I had to switch to `@yao-pkg/pkg`. I tested the plugin on Linux and Windows.
I also tested to build and start as a remote server and to build and run the Docker image. Everything works fine. 

Fixes https://github.com/grafana/grafana-image-renderer/issues/518

PR #619, opened 2025-03-27
Chore: Downgrade to Node 20

evictorero:
We [upgraded to node 22](https://github.com/grafana/grafana-image-renderer/pull/595) because of node 18 support ends in April. We also moved to [yao-pkg](https://github.com/yao-pkg) for packaging because [pkg](https://github.com/vercel/pkg) is deprecated.
Since there are some issues with [node 22](https://github.com/yao-pkg/pkg/issues/87) packaging and Windows we decided to downgrade to node 20 LTS which has support [until April 2026](https://nodejs.org/en/about/previous-releases#release-schedule).


nmarrs (approved):
LGTM - just confirming we are keeping the Yao-pkg change as part of this right? 

Great work! 💫

evictorero:
> LGTM - just confirming we are keeping the Yao-pkg change as part of this right?
> 
> Great work! 💫

Yes, we keep `yao-pkg`.

PR #628, opened 2025-04-17
PDF: Use sent timeout

AgnesToulet:
The `.pdf` function currently uses the default timeout (30s) instead of the provided timeout as every other function during the rendering process.

When printing a PDF for a very large dashboard, this leads to not being able to customize the timeout.

### DIFF
--- a/src/browser/browser.ts
+++ b/src/browser/browser.ts
@@ -381,6 +381,7 @@ export class Browser {
           });
         }
 
+        const timeoutMs = options.timeout * 1000;
         return page.pdf({
           ...getPDFOptionsFromURL(options.url),
           margin: {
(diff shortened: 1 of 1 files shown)


src/browser/browser.ts
```diff
@@ -391,6 +391,7 @@ export class Browser {
           },
           path: options.filePath,
           scale: 1 / scale,
+          timeout: options.timeout * 1000,
```
lucychen-grafana, line 394: where is the 1000 coming from?

nmarrs: my guess is that this option is in passed as number of seconds and the option needs to be converted to `ms` 

we could break this out into a separate const to make this more clear :)

evictorero (approved):

PR #633, opened 2025-04-29
Load tests: Improve scripts

AgnesToulet:
This PR fixes and improves our load tests:
- Fix login workflow
- Add support for service account token
- Use new k6 Docker image

+ some clean up in Docker files (as the version is no longer used and generate a warning)

### DIFF
--- a/devenv/loadtest/README.md
+++ b/devenv/loadtest/README.md
@@ -21,12 +21,6 @@ Run load test for custom duration:
 $ ./run.sh -d 10s
 ```
 
-Run only 1 iteration of the load test (useful for testing):
-
-```bash
-$ ./run.sh -i 1
-```
-
 Run load test for custom target url:
 
 ```bash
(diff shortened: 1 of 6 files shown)


devenv/loadtest/README.md
```diff
@@ -24,6 +25,7 @@ Run only 1 iteration of the load test (useful for testing):
```
lucychen-grafana: Given VU=2, and the number of iterations can't be less than the number of VUs, this doesn't work/may be confusing. Maybe we can describe it as `Run load test for custom iteration (useful for testing):`

AgnesToulet: Good catch, I updated the doc :+1:

devenv/loadtest/modules/client.js
```diff
@@ -1,4 +1,4 @@
-import http from "k6/http";
+import http from 'k6/http';
```
nmarrs, line 1: Do we have linting / prettier set up for this repo? 🤔

nmarrs (approved):
Code changes look good - didn't test locally though so @lucychen-grafana it be a good idea for you to take another look at this one :)

devenv/loadtest/modules/client.js
AgnesToulet: We don't have a check in CI to enforce it, hence the issues that are randomly fixed when people update the code with prettier set :sweat_smile:

PR #640, opened 2025-05-13
Tracing: Fix tracing headers causing CORS issue

AgnesToulet:
It seems that `page.on('request', ...)` isn't triggered on EVERY request so adding the headers for every request and removing them here wasn't working for some requests.

This is achieving the same result, the other way around, only adding headers for requests we are sure we should add them.

### DIFF
--- a/src/browser/browser.ts
+++ b/src/browser/browser.ts
@@ -568,7 +568,7 @@ export class Browser {
     return callback();
   }
 
-  async addPageListeners(page: puppeteer.Page, headers) {
+  async addPageListeners(page: puppeteer.Page, headers: HTTPHeaders) {
     page.on('error', this.logError);
     page.on('pageerror', this.logPageError);
     page.on('requestfailed', this.logRequestFailed);
@@ -627,7 +627,7 @@ export class Browser {
           headers['tracestate'] = optionsHeaders['tracestate'] ?? '';
         }
       } catch (error) {
-        this.log.debug('Failed to remove tracing header', 'url', url, 'referer', referer, 'error', error.message);
+        this.log.debug('Failed to add tracing headers', 'url', url, 'referer', referer, 'error', error.message);
       }
 
       req.continue({ headers });
--- a/src/browser/browser.ts
+++ b/src/browser/browser.ts
@@ -8,7 +8,7 @@ import * as promClient from 'prom-client';
 import * as Jimp from 'jimp';
 import { Logger } from '../logger';
 import { RenderingConfig } from '../config/rendering';
-import { ImageRenderOptions, RenderOptions } from '../types';
+import { HTTPHeaders, ImageRenderOptions, RenderOptions } from '../types';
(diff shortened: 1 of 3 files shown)


src/browser/browser.ts
```diff
@@ -168,9 +168,11 @@ export class Browser {
       });
     }
 
-    if (options.headers && Object.keys(options.headers).length > 0) {
-      this.log.debug(`Setting extra HTTP headers for page`, 'headers', options.headers);
-      await page.setExtraHTTPHeaders(options.headers as any);
+    if (options.headers && options.headers['Accept-Language']) {
+      const headers = { 'Accept-Language': options.headers['Accept-Language'] };
+      this.log.debug(`Setting extra HTTP headers for page`, 'headers', headers);
```
nmarrs, line 173: do we want to keep these debug logs for future troubleshooting?

nmarrs (approved):
LGTM

evictorero (approved):

src/browser/browser.ts
```diff
@@ -168,9 +168,11 @@ export class Browser {
       });
     }
 
-    if (options.headers && Object.keys(options.headers).length > 0) {
-      this.log.debug(`Setting extra HTTP headers for page`, 'headers', options.headers);
-      await page.setExtraHTTPHeaders(options.headers as any);
+    if (options.headers && options.headers['Accept-Language']) {
```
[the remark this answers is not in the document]
AgnesToulet: It was how it worked before, we were transferring only this header from the `options`: https://github.com/grafana/grafana-image-renderer/pull/612/files#diff-ee2c77c792e3531313c7228565dfc561d43a581fdf2730ff266c0ef17b13d43fL61-L69

I updated it to keep tracing headers but the behavior stays the same.

src/browser/browser.ts
```diff
@@ -592,38 +594,44 @@ export class Browser {
...
+        const refererHostname = referer ? new URL(referer).hostname : '';
+        const shouldAddHeaders = req.isNavigationRequest() || urlHostname === refererHostname;
+        this.log.debug('Comparing referer and URL hostnames', 'method', method, 'shouldAddHeaders', shouldAddHeaders, 'url', url, 'referer', referer);
+
+        if (shouldAddHeaders) {
+          headers['traceparent'] = optionsHeaders['traceparent'] ?? '';
+          headers['tracestate'] = optionsHeaders['tracestate'] ?? '';
+        }
+      } catch (error) {
+        this.log.debug('Failed to remove tracing header', 'url', url, 'referer', referer, 'error', error.message);
```
evictorero, line 630: Good catch from Copilot
```suggestion
        this.log.debug('Failed to add tracing headers', 'url', url, 'referer', referer, 'error', error.message);
```

src/browser/browser.ts
AgnesToulet: I didn't add this one but yes I think it can be interesting for troubleshooting.

PR #661, opened 2025-07-16
Docker: Move to distroless Debian

Proximyst:
Incident: https://ops.grafana-ops.net/a/grafana-irm-app/incidents/3491


src/service/http-server.ts
```diff
@@ -130,8 +130,9 @@ export class HttpServer {
 
       try {
         const browserVersion = await this.browser.getBrowserVersion();
+        this.log.info('Browser version', 'version', browserVersion);
         browserInfo.labels(browserVersion).set(1);
-      } catch {
+      } catch (err) {
```
macabu: Do we want to keep these changes? The `err` is not being used

Dockerfile
```diff
@@ -1,71 +1,53 @@
...
 
-# Production dependencies stage
-FROM base AS prod-dependencies
-
-COPY package.json yarn.lock ./
-RUN yarn install --pure-lockfile --production
-
-# Final stage
-FROM base
+FROM gcr.io/distroless/nodejs22-debian12:debug-nonroot
```
macabu: `debug` variant intentional?

macabu (commented):
First pass

Dockerfile
Proximyst: Yep; it has some binaries we need.

macabu (approved):

roock:
@Proximyst There are instructions in the official Grafana Docs about how to add custom root certificates to the grafana renderer which do not work with distroless debian anymore:
https://grafana.com/docs/grafana/latest/setup-grafana/image-rendering/troubleshooting/#certificate-signed-by-internal-certificate-authorities
It would be very helpful if the certutil could be part of the official image, that would make things much easier to add a custom certificate.

Proximyst:
@roock Thanks for pointing that out! We'll take a look at implementing something similar in the coming days/weeks. Sorry for the impact this has caused you 😓

PR #690, opened 2025-07-28
Dockerfile: remove fixed UID

ioachim-hub:
while at it:
- adapt Dockerfile to work building using Podman

The rationale stands in letting the user specify the user ID on k8s SecurityContext configuration.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM debian:12-slim AS debian-updated
+FROM docker.io/debian:12-slim AS debian-updated
 
 SHELL ["/bin/bash", "-euo", "pipefail", "-c"]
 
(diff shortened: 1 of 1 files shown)


ioachim-hub:
@Proximyst
Please take a look if you agree with these changes

Dockerfile
```diff
@@ -66,7 +66,6 @@ EXPOSE 8081
 
 # Simple regression test for: https://github.com/grafana/grafana-image-renderer/issues/686
 RUN test "$(id -u)" -eq 65532
-USER 65532
```
Proximyst, line 69: This is required for containerd-based Kubernetes. If Podman can't support it, we won't support it.

Proximyst (changes requested):

Dockerfile
ioachim-hub: This is not because of the podman.
The other changes are related to podman.

This line is removed because it sets a fixed user ID which is not a good practice.
Few lines above, there is set `USER nonroot` which respects the containerd-based kubernetes requirements.

ioachim-hub: Tested it using these configuration:
```
      securityContext:
        runAsUser: 65534
        runAsGroup: 65534
        runAsNonRoot: true
```

Proximyst: We set the numeric UID as per a requirement in Kubernetes: if you use `runAsNonRoot: true` without setting a UID manually (rather trusting the Dockerfile), you will get an error. We would prefer you don't hardcode a UID in your deployments, instead trusting us + that non-root option to set the service up properly.

As such, we are not going to drop this directive.

Proximyst:
Closing as rejected; see PR line comment.

PR #737, opened 2025-08-19
feat: define basic Go application

Proximyst:
This starts the move towards writing the application in Go. This targets the service mode only; plugin mode will come in a later PR.

This makes no effort at replicating _all_ the behaviours found in the Node service, but it gets us to a position where we have a basis to migrate.

_Why Go?_ See the ADR :)
_Why these metrics/logs?_ See the ADR :)
_Why ADRs?_ See the ADR :)

Ticket: https://github.com/grafana/grafana-operator-experience-squad/issues/1514


pkg/api/middleware/trustedUrl.go
```diff
@@ -0,0 +1,30 @@
+package middleware
```
macabu: Nit: I think the file name could be `trusted_url`

pkg/api/middleware/trustedUrl.go
```diff
@@ -0,0 +1,30 @@
...
+
+var MetricTrustedURLRequests = prometheus.NewCounterVec(prometheus.CounterOpts{
+	Name: "http_trusted_url_requests",
+	Help: "Counts the requests with URL queries",
+}, []string{"result"})
+
+// TrustedURL ensures that the `url` query parameter (if it exists) aims at an HTTP or HTTPS website, disallowing e.g. `chrome://` and `file://`.
+func TrustedURL(h http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		url := r.URL.Query().Get("url")
```
macabu: Should we `url.Parse` and check the `Scheme` to avoid any shenanigans?

pkg/api/version.go
```diff
@@ -0,0 +1,26 @@
...
+
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		version, err := browser.GetVersion(r.Context())
+		if err != nil {
+			http.Error(w, "failed to get browser version", http.StatusInternalServerError)
+			slog.ErrorContext(r.Context(), "failed to get browser version", "error", err)
+			return
+		}
+
+		_, _ = w.Write([]byte("grafana-image-renderer " + serviceVersion + "\n"))
```
macabu, line 23: Why this and not JSON for example? What is consuming this data?

pkg/api/middleware/metrics.go
```diff
@@ -0,0 +1,34 @@
...
+		Help: "How many requests are in flight?",
+	})
+	MetricRequestDurations = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Name: "http_request_duration",
+		Help: "How long does a request take?",
+		ConstLabels: prometheus.Labels{
+			"unit": "seconds",
+		},
+		Buckets: []float64{0.5, 1, 3, 4, 5, 7, 9, 10, 11, 15, 19, 20, 21, 24, 27, 29, 30, 31, 35, 55, 95, 125, 305, 605},
+	}, []string{"method", "path"}) // TODO: Output HTTP status codes, too
```
macabu, line 22: I suppose something like this could work for getting the status codes https://github.com/grafana/grafana-cloud-migration-service/blob/main/pkg/server/middleware/middleware.go

adrs/0001-language.md
```diff
@@ -0,0 +1,26 @@
...
+
+* The new owners, Grafana Enterprise, is largely Go-first engineers who hire as such.
+* The organisation's backend engineers at large primarily know Go rather than Node & TS.
+* Other languages would open for more powerful security features that are inherently incompatible with how Node.js v22 works.
+
+## Considered Options
+
+* Do nothing: continue using Node.js.
+* Use Go: a native, simple language we already have a pretty vast amount of experience with.
+* Use Rust: a native, complex language that is very, very powerful and has a great ecosystem with strong typing, however is hard to hire for and has little adoption internally.
```
macabu, line 18: So close yet so far

pkg/api/version.go
Proximyst: Nothing :) I wanted a quick first step way to actually test we can use the Chromium instance at all

pkg/api/middleware/trustedUrl.go
Proximyst: good idea! doing that now :)

macabu (approved):

PR #738, opened 2025-08-19
feat(go): support config files

Proximyst:
This will now start reading a `config.yaml` (and `config.yml`, and `config.json`) to gather arguments that aren't passed in on the CLI.


cmd/config/cmd.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/urfave/cli/v3"
+	"gopkg.in/yaml.v3"
+)
+
+func NewCmd() *cli.Command {
+	return &cli.Command{
+		Name:  "config",
+		Usage: "Display the configuration of the service.",
+		Description: `Read and recreate a merged, single view configuration file from the files used by the application.
+This reads the config.yaml, config.yml, and config.json files, if they exist.
```
macabu, line 18: Is the intention that we'd support multiple configs with different formats? Generally I think that could be confusing for the user to have to know which one takes precedence and whatnot and would only want to have one of them at a time. Meaning instead of merging the configs we could return an error

cmd/config/source.go
```diff
@@ -0,0 +1,16 @@
+package config
+
+import (
+	altsrc "github.com/urfave/cli-altsrc/v3"
+	"github.com/urfave/cli-altsrc/v3/yaml"
+	"github.com/urfave/cli/v3"
+)
+
+func FromConfig(path string) cli.ValueSourceChain {
+	return cli.NewValueSourceChain(
```
macabu, line 10: Does this unmarshal at least one config file every time we try to read a value from it? Or does it have some memory?

cmd/config/cmd.go
Proximyst: No, this is just how the altsrc library works, so it's recreated here :)

cmd/config/source.go
Proximyst: I believe it reads the config values over and over when starting up. I'd argue that's totally OK; it's not expected to have a massive config, nor is it expected to launch the service on every request

macabu (approved):

PR #741, opened 2025-08-25
feat(go): support rendering Grafana dashboards

Proximyst:
This extends the rendering support to work for Grafana dashboards.

### DIFF
--- a/pkg/chromium/browser.go
+++ b/pkg/chromium/browser.go
@@ -96,7 +96,7 @@ func (r *RenderingOptions) SetHeader(key, value string) {
 
 func (r *RenderingOptions) Normalise(prototype RenderingOptions) error {
 	if strings.HasPrefix(r.URL, "socket://") {
-		return fmt.Errorf("Image rendering in socket mode is not supported")
+		return fmt.Errorf("image rendering in socket mode is not supported")
 	}
 
 	if r.Timeout <= 0 {
--- a/pkg/chromium/browser.go
+++ b/pkg/chromium/browser.go
@@ -9,6 +9,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/chromedp/cdproto/emulation"
+	"github.com/chromedp/cdproto/network"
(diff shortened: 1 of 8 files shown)


pkg/chromium/browser.go
```diff
@@ -176,3 +302,105 @@ func printPDF(requestOptions RenderingOptions, dst chan []byte) chromedp.Action
...
+		for range scrolls {
+			err := chromedp.Evaluate(`window.scrollBy(0, window.innerHeight, { behavior: 'instant' })`, nil).Do(ctx)
+			if err != nil {
+				return fmt.Errorf("failed to scroll: %w", err)
+			}
+			time.Sleep(timeBetweenScrolls)
+		}
+
+		err = chromedp.Evaluate(`window.scrollTo(0, 0, { behavior: 'instant' })`, nil).Do(ctx)
+		if err != nil {
```
macabu, line 338: Why do we need to scroll back to the top? Would it be possible to write a comment explaining?

pkg/chromium/browser.go
```diff
@@ -69,15 +83,80 @@ type RenderingOptions struct {
...
-// RenderPDF visits the website and waits for it to be fully rendered.
+func (r *RenderingOptions) SetHeader(key, value string) {
+	if r.Headers == nil {
+		r.Headers = make(map[string]string)
+	}
+	r.Headers[key] = value
+}
+
+func (r *RenderingOptions) Normalise(prototype RenderingOptions) error {
+	if strings.HasPrefix(r.URL, "socket://") {
```
macabu, line 98: Do we want to support it at some point?

macabu (approved):

pkg/chromium/browser.go
Proximyst: Not a clue! This is just what the TS service does.

pkg/chromium/browser.go
Proximyst: Probably not. Again, TS does this, so I just copied the behaviour.

PR #776, opened 2025-09-12
fix(go): bug in CSV rendering

Proximyst:
this doesn't have any fix yet, just a draft PR. this reproduces the bug, but I have no clue why it happens yet.


Proximyst:
ack, we've decided not to care :)

PR #783, opened 2025-09-16
test(acceptance): ensure geomap panels work

Proximyst:
This adds two new dashboards, with geomap panels. These show us that the panels work with our (default) service settings.

There are more commits here that were effectively removed by merging with main :) I split some parts off to #788 & #787, so this PR becomes a single unit instead.

Related: https://github.com/grafana/support-escalations/issues/18419


devenv/docker/go-build/dashboards
Proximyst, line 1: _Why symlink?_ We can now put the dashboards in _one_ place rather than two, reducing risk of difference and ensuring we see the exact test env in our devenv.

leandro-deveikis (approved):

PR #802, opened 2025-10-10
feat: add more tracing to requests

Proximyst:
Mostly to try to figure out what's going on with requests timing out before the configured timeout in dev 🤔


Proximyst:
split off into many PRs

PR #805, opened 2025-10-13
feat(go): log requests on info with duration

Proximyst:
We should be logging more at info level, so that we have a generally good idea of what's up. This is more useful than searching traces for specific attributes in quick validation testing.

### DIFF
--- a/pkg/api/middleware/logger.go
+++ b/pkg/api/middleware/logger.go
@@ -4,18 +4,21 @@ import (
 	"log/slog"
 	"net/http"
 	"sync"
+	"time"
 )
 
 func RequestLogger(h http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		start := time.Now()
 		lw := &loggingResponseWriter{w: w}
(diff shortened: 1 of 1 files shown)


pkg/api/middleware/logger.go
```diff
@@ -4,18 +4,21 @@ import (
...
+	"time"
 )
 
 func RequestLogger(h http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		start := time.Now()
 		lw := &loggingResponseWriter{w: w}
 		defer func() {
-			slog.DebugContext(r.Context(), "request complete",
+			slog.InfoContext(r.Context(), "request complete",
```
tskarhed, line 15: Will we switch this back to Debug when we've validated?

Proximyst: Nah, I don't think so. We don't get the type of request volume in this service that we should care all too much about the logging volume, rather it's nice to optimise for handiness (and otherwise, we can move to WARN+ in prod, which is also totally fine).

PR #807, opened 2025-10-13
feat(go): add more tracing details

Proximyst:
This has helped me a lot in testing in dev. We get a whole lot more details in the spans for everything now.

### DIFF
--- a/pkg/api/middleware/logger.go
+++ b/pkg/api/middleware/logger.go
@@ -4,18 +4,21 @@ import (
 	"log/slog"
 	"net/http"
 	"sync"
+	"time"
 )
 
 func RequestLogger(h http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		start := time.Now()
 		lw := &loggingResponseWriter{w: w}
 		defer func() {
-			slog.DebugContext(r.Context(), "request complete",
+			slog.InfoContext(r.Context(), "request complete",
 				"method", r.Method,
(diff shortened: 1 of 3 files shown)


pkg/api/render.go
```diff
@@ -37,39 +41,52 @@ func HandleGetRender(browser *service.BrowserService) http.Handler {
...
+			span.SetStatus(codes.Error, "url query param empty")
 			http.Error(w, "missing 'url' query parameter", http.StatusBadRequest)
 			return
 		}
 		targetURL, err := url.Parse(rawTargetURL)
 		if err != nil {
+			span.SetStatus(codes.Error, "url query param was unparseable")
+			span.RecordError(err, trace.WithAttributes(attribute.String("url", rawTargetURL)))
 			http.Error(w, fmt.Sprintf("invalid 'url' query parameter: %v", err), http.StatusBadRequest)
 			return
```
tskarhed, line 53: Why isn't the usual Go pattern with returning the error used here? Is it a typical thing to write to the http handler instead? My main thinking is that you could just return a generic error, then set the status directly based on that. I guess it would be a bit harder to attach the error attributes then anyways

Proximyst: HTTP handlers in Go are messy in all cases... it's something I've been meaning to improve on, but this works well enough, so I haven't really been bothered. I'm intending on trying out [Gin](https://gin-gonic.com/), [Fiber](https://gofiber.io/), and [Echo](https://echo.labstack.com/) as well, to see if they can be of any particular use for us here.

PR #820, opened 2025-10-20
feat(go): track chromium memory usage

Proximyst:
This is the first step to rate-limiting smartly based on the amount of browsers we have running.

### DIFF
--- a/tests/acceptance/fixtures/dashboards/grafana-image-renderer.json
+++ b/tests/acceptance/fixtures/dashboards/grafana-image-renderer.json
@@ -878,7 +878,7 @@
         "type": "prometheus",
         "uid": "${prom}"
       },
-      "description": "This is the peak memory usage reported by each process on every request.",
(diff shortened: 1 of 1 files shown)


pkg/service/process.go
```diff
@@ -0,0 +1,130 @@
...
+		peakMemory := 0
+		defer func() {
+			// We only do the lock once per process. This reduces contention significantly.
+			p.mu.Lock()
+			defer p.mu.Unlock()
+
+			if p.PeakMemory == 0 {
+				p.PeakMemory = int64(peakMemory)
+			} else {
+				const decay = 5
```
macabu, line 102: Would we want the decay to be configurable by flag?

macabu (approved):

pkg/service/process.go
Proximyst: Yeah, probably. I'll do that later :)

PR #823, opened 2025-10-21
feat(go): add dynamic rate-limiting

Proximyst:
This adds rate-limiting to the service. The rate-limiter is dynamic, based on the memory usage of the Chromium processes; this is the first bottleneck that servers will generally hit as it is also one that kills the server.

The bulk of the diff numbers come from the dashboards.

### DIFF
--- a/pkg/api/middleware/ratelimiter.go
+++ b/pkg/api/middleware/ratelimiter.go
@@ -60,7 +60,7 @@ func (p processBasedLimiter) Limit(next http.Handler) http.Handler {
 
 			w.Header().Set("Retry-After", "5")
 			w.WriteHeader(http.StatusTooManyRequests)
-			w.Write([]byte("server is too busy, try again later"))
+			_, _ = w.Write([]byte("server is too busy, try again later"))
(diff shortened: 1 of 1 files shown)


pkg/config/config.go
```diff
@@ -588,3 +588,114 @@ func BrowserConfigFromCommand(c *cli.Command) (BrowserConfig, error) {
...
+				if i < 1 {
+					return fmt.Errorf("rate-limit.process-tracker.decay must be at least 1")
+				}
+				return nil
+			},
+		},
+		&cli.DurationFlag{
+			Name:    "rate-limit.process-tracker.interval",
+			Usage:   "How often to sample process statistics on the browser processes. Must be >= 1ms.",
+			Value:   50 * time.Millisecond,
```
macabu, line 648: The previous default was 500, is the change here intentional?

macabu (approved):

pkg/config/config.go
Proximyst: yep!

PR #828, opened 2025-10-24
fix(go): return correct CSV file names

Proximyst:
Thanks to @liamoddell for finding & reporting this ❤️ 

Test example:
image

### DIFF
--- a/pkg/api/rendercsv.go
+++ b/pkg/api/rendercsv.go
@@ -5,7 +5,10 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"mime"
 	"net/http"
+	"path"
+	"path/filepath"
 	"strconv"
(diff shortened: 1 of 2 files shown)


pkg/api/rendercsv.go
```diff
@@ -97,8 +100,24 @@ func HandleGetRenderCSV(browser *service.BrowserService) http.Handler {
 		MetricRenderCSVDuration.WithLabelValues("success").Observe(time.Since(start).Seconds())
 		span.SetStatus(codes.Ok, "csv rendered successfully")
 
+		requestedFilePath := r.URL.Query().Get("filePath")
```
tskarhed, line 103: How is this different to the `fileName`? Do we ignore cases where this is a path rather than just the filename?

PR #830, opened 2025-10-27
feat(go): abstract away the browser

Proximyst:
This is not ready for review; the intent of this is to figure out a way to potentially offer more browser options (such that we can test them out proper). My primary candidates for testing are currently these:

* [Firefox](https://www.firefox.com/en-US/) (via [GeckoDriver](https://github.com/mozilla/geckodriver)?) (MPL-2.0)
* [Microsoft Edge](https://www.microsoft.com/en-us/edge/)? (Proprietary, closed-source; unsure if we can redistribute)
* [Ladybird](https://ladybird.org/) (BSD-2-Clause)
* [Lightpanda](https://lightpanda.io/) (AGPL-3.0)

It's not entirely clear whether we could really ship all of these, but we could at least _support_ them.


Proximyst:
closing down; we aren't going to earn much from this without a proper BiDi lib :(

PR #835, opened 2025-10-29
feat(go): run Chromium in Linux namespaces

Proximyst:
This creates new Linux namespaces for Chromium to run in.

It won't solve _all_ our issues, but is a good start.

image


cmd/sandbox/cmd.go
```diff
@@ -0,0 +1,204 @@
...
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"syscall"
+
+	"github.com/grafana/grafana-image-renderer/pkg/sandbox"
+	"github.com/urfave/cli/v3"
+	"go.opentelemetry.io/otel/trace"
+	libcap "kernel.org/pub/linux/libs/security/libcap/cap"
```
macabu, line 16: TIL about this pkg

PR #864, opened 2025-11-13
feat: use URL type for tracing addresses

mpldr:
This simplifies the logic a bit and is less fragile when some tracing backends have odd connection strings.

### DIFF
--- a/pkg/traces/traces.go
+++ b/pkg/traces/traces.go
@@ -6,8 +6,8 @@ import (
 	"crypto/x509"
 	"fmt"
 	"log/slog"
+	"net/url"
 	"os"
-	"strings"
 
 	"github.com/grafana/grafana-image-renderer/pkg/config"
 	"go.opentelemetry.io/otel/exporters/otlp/otlptrace"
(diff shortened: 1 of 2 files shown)


pkg/traces/traces.go
```diff
@@ -74,31 +74,23 @@ func NewTracerProvider(ctx context.Context, cfg config.TracingConfig) (*sdktrace
 	}
 
 	var exporter *otlptrace.Exporter
-	if strings.HasPrefix(cfg.Endpoint, "http://") || strings.HasPrefix(cfg.Endpoint, "https://") {
+	dsn, err := url.Parse(cfg.Endpoint)
```
macabu, line 77:
```suggestion
	dsn, err := url.Parse(cfg.Endpoint)
	if err != nil {
		return nil, fmt.Errorf("failed to parse tracing endpoint %q: %w", cfg.Endpoint, err)
	}
```

pkg/traces/traces.go
```diff
@@ -115,7 +107,6 @@ func NewTracerProvider(ctx context.Context, cfg config.TracingConfig) (*sdktrace
 		}
 	} else { // gRPC
 		slog.InfoContext(ctx, "setting up gRPC trace exporter", "endpoint", cfg.Endpoint)
-		_, cfg.Endpoint, _ = strings.Cut(cfg.Endpoint, "://")
 
 		var opts []otlptracegrpc.Option
 		opts = append(opts, otlptracegrpc.WithEndpoint(cfg.Endpoint))
```
macabu, line 112:
```suggestion
		opts = append(opts, otlptracegrpc.WithEndpointURL(dsn.String()))
```

macabu (changes requested):
Hi @mpldr , thanks for the PR and sorry for the delay!

I tried applying the changes requested myself but didn't have permission to push to your fork.

Could you also merge/rebase with `master` so it can run the latest CI changes? Thanks

PR #880, opened 2025-12-09
Add config for waiting for multiple successful cycles until rendering

mmandrus:
Active queries can drop down 0 before before incrementing again. If there is enough of a pause between these two events, the dashboard can be rendered before it is ready. We are seeing that on [this escalation](https://github.com/grafana/support-escalations/issues/19470), where the active queries temporarily drop to 0 for 40ms at times, causing the dashboard to render blank.

Starting out implementing this as a global setting so that we can get a fix out to customers sooner. Once we have validated it, we can change it to a request-specific config.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -322,8 +322,8 @@ type BrowserConfig struct {
 	// ReadinessTimeout is the maximum time to wait for the web-page to become ready (i.e. no longer loading anything).
 	ReadinessTimeout           time.Duration
 	ReadinessIterationInterval time.Duration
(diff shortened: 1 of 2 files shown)


pkg/config/config.go
```diff
@@ -447,6 +449,17 @@ func BrowserFlags() []cli.Flag {
 				return nil
 			},
 		},
+		&cli.IntFlag{
+			Name:  "browser.readiness.wait-for-n-cycles",
```
Proximyst: i'd name this something more like `wait-for-n-query-cycles` as it only affects queries

mmandrus: done [111ca10](https://github.com/grafana/grafana-image-renderer/pull/880/commits/111ca109e83fa0b427df1127913313488dede0aa)

lucychen-grafana (approved):

PR #885, opened 2025-12-11
Add support for overriding first query timeout for specific slugs

mmandrus:


pkg/config/config.go
```diff
@@ -483,6 +487,19 @@ func BrowserFlags() []cli.Flag {
 			Value:   time.Second * 3,
 			Sources: FromConfig("browser.readiness.give-up-on-first-query", "BROWSER_READINESS_GIVE_UP_ON_FIRST_QUERY"),
 		},
+		&cli.StringMapFlag{
```
Proximyst, line 490: this must be `Hidden: true` so we don't make it official API

mmandrus:
closing to build a more robust/multi-purpose solution per discussion with Mariell

PR #898, opened 2026-01-07
Example configuration files

maxlemieux:
Complete example configuration files showing the default values for all settings. We had this in v3, and example files would also be helpful for v5.

### DIFF
--- a/default.json
+++ b/default.json
@@ -1,0 +1,64 @@
+{
+  "log": {
+    "level": "info"
+  },
+  "server": {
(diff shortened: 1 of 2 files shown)


maxlemieux:
Looking back, we only had JSON, understanding it is equivalent to YAML maybe we don't need the YAML one.

maxlemieux:
What is the preferred generation approach? I would be happy to look into it, to improve this PR.

PR #909, opened 2026-01-16
API: Add config flag for default encoding in rendering endpoint

macabu:
Previously if you didn't pass an `encoding` query param in the `/render` request, the default would be a hardcoded `pdf`.

This adds yet another setting to allow users to change this default to `png` as well.


tests/acceptance/rendering_grafana_test.go
```diff
@@ -271,7 +271,6 @@ func TestRenderingGrafana(t *testing.T) {
 			req.Header.Set("X-Auth-Token", "-")
 			query := req.URL.Query()
 			query.Set("url", "http://grafana:3000/d/provisioned-prom-testing?render=1&from=1699333200000&to=1699344000000&kiosk=true")
-			query.Set("encoding", "pdf")
```
macabu, line 274: I wasn't really keen on adding two new acceptance tests for this, so I remove the encoding query param since the test name is already "with defaults", so it matched the theme :)

pkg/config/config.go
```diff
@@ -76,6 +76,35 @@ func (l LogLevel) ToSlog() (slog.Leveler, error) {
 	}
 }
 
+type APIConfig struct {
```
tskarhed, line 79: This naming is because it configures the behavior of the API, rather than the browser, right?

macabu: Exactly!

PR #923, opened 2026-02-02
ci: sign output images and manifests

Proximyst:
This implements Cosign signatures for OCI images and manifests. We want this to ensure that we can validate that the software we run is _ours_ and built on GitHub Actions.

With GH built image:

```console
$ cosign verify --certificate-oidc-issuer=https://token.actions.githubusercontent.com --certificate-identity-regexp='https://github\\.com/grafana/[^/]+/.github/workflows/.*' (crane digest --full-ref ghcr.io/grafana/grafana-image-renderer:dev-pull-923-5cb9dc7a41976c6fc7d80cb89e12f78b5d77e01b)

Verification for ghcr.io/grafana/grafana-image-renderer@sha256:adf3323d8cd5f8f5d3e4e884351f780b2382e4f2b21adefde1bb624d7b611b30 --
The following checks were performed on each of these signatures:
  - The cosign claims were validated
  - Existence of the claims in the transparency log was verified offline
  - The code-signing certificate was verified using trusted certificate authority certificates

[{"critical":{"identity":{"docker-reference":"ghcr.io/grafana/grafana-image-renderer@sha256:adf3323d8cd5f8f5d3e4e884351f780b2382e4f2b21adefde1bb624d7b611b30"},"image":{"docker-manifest-digest":"sha256:adf3323d8cd5f8f5d3e4e884351f780b2382e4f2b21adefde1bb624d7b611b30"},"type":"https://sigstore.dev/cosign/sign/v1"},"optional":{}}]
```

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -119,6 +119,7 @@ jobs:
       attestations: write # required to create attestations for the built image
       id-token: write # required to create attestations for the built image, and to read secrets
       pull-requests: write # required to comment on the pull request
+      artifact-metadata: write # to generate artifact metadata records; action silently skips if this is not given
     steps:
       - name: Log into GHCR
         if: needs.tag.outputs.registry == 'ghcr.io'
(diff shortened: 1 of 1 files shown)


psof (approved):
Solid producer step for keyless signing, SHA-pinned actions, and full artifact-graph coverage (children + manifest + :latest) are exactly what we want, and no static keys means nothing to rotate or leak.

One thing to flag before any admission policy leans on this: the verify identity in the description,`grafana/[^/]+/.github/workflows/.*`, is an org-wide wildcard, not a trust boundary. Signing from this repo's own `build.yaml` is fine to bootstrap, but if we pin admission on a per-repo identity, the boundary becomes "any workflow in any grafana repo." Strongly prefer moving the cosign sign logic into a `grafana/shared-workflows` reusable workflow so the Fulcio SAN is a single pinnable org-wide identity, you already consume shared-workflows here so it's a short hop.

Two smaller points: signing is unconditional across events, so dev-pull-* PR images are signed by the same identity as releases the signature attests "built on GHA," not "built from a reviewed ref," so we'll need to distinguish those before enforcement.

Net: ship it as a producer enablement; just don't let the per-repo identity or the wildcard regexp harden into the admission contract.

PR #930, opened 2026-02-19
Reporting: Use performance utils to decide when a dashboard is done

tskarhed:
Attaches a method to the window object, which can be called by the frontend when the event is finished. Requires the changes in https://github.com/grafana/grafana/pull/118312. It check for a global flag on the window object for support for the binding.

### DIFF
--- a/pkg/service/browser.go
+++ b/pkg/service/browser.go
@@ -1081,7 +1081,6 @@ func waitForReady(browserCtx context.Context, cfg config.BrowserConfig, url stri
 			span.RecordError(err)
 		}
 
-
 		if supportsBinding {
 			MetricBrowserReadinessMode.WithLabelValues("binding").Inc()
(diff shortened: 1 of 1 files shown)


pkg/service/browser.go
```diff
@@ -1060,6 +1075,34 @@ func waitForReady(browserCtx context.Context, cfg config.BrowserConfig, url stri
 			readinessTimeout = time.After(requestConfig.ReadinessTimeout)
 		}
 
+		var supportsBinding = false // default to legacy polling
+		// Set in core Grafana
+		if err := chromedp.Evaluate(`!!window.__grafanaRenderBindingSupported`, &supportsBinding).Do(ctx); err != nil {
+			span.RecordError(err)
+		}
+
```
macabu, line 1083: Incredibly tiny nit, extra newline
```suggestion
```

PR #932, opened 2026-02-20
Add TLS configuration to server listener

mrevutskyi:
Currently server always uses HTTP even if SSL certs are provided, becaue default HTTP listner gets created at the beginning, and no TLS on top of it

Test:
```
./grafana-image-renderer server  --server.cert identity.cert --server.key identity.key
```

before:
```
curl -k https://localhost:8081/healthz
curl: (35) OpenSSL/1.1.1k: error:1408F10B:SSL routines:ssl3_get_record:wrong version number
```

after
```
curl -k "https://localhost:8081/healthz"
OK
```


macabu (approved):
Thank you for the contribution!

PR #979, opened 2026-05-07
feat: bump all Go dependencies

Proximyst:
Bumps the Go deps, as we haven't done so in a bit. We should try to figure out why Renovate/Dependabot doesn't do this for us.


macabu (dismissed):
I guess the cdproto version is not compatible with our fork, otherwise LGTM

macabu (approved):

PR #1016, opened 2026-07-10
Docker: Add Debian distroless variant (experimental)

macabu:
Testing... WIP


macabu:
Still has too many CVEs, largely due to the direct dependencies of Chromium, so wouldn't be very useful with the added maintenance burden

PR #1041, opened 2026-08-28
API: harden healthz response handling

Alja9:
Hi Team,

This PR for improve healthz endpoint error handling with explicit logging and OpenTelemetry span recording for better observability.

The handler now logs write failures instead of silently ignoring them, making it easier to debug unexpected issues like broken pipes or network timeouts.

go test `./pkg/api/api/` ok


pkg/api/healthz.go
```diff
@@ -1,14 +1,23 @@
...
+	"log/slog"
+)
 
 func HandleGetHealthz() http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		tracer := tracer(r.Context())
 		_, span := tracer.Start(r.Context(), "HandleGetHealthz")
 		defer span.End()
 
+		w.Header().Set("X-Content-Type-Options", "nosniff")
```
macabu: I don't believe this one is strictly necessary as 1) traffic should always be internal and trusted, 2) requests to this endpoint are not done by a browser

pkg/api/healthz.go
```diff
@@ -1,14 +1,23 @@
...
+		w.Header().Set("Cache-Control", "no-cache, no-store")
```
macabu: I don't think this is needed either, I don't believe tools like Kubernetes or Docker or cURL cache the response, and this is not being queried by a browser ideally.

pkg/api/healthz.go
```diff
@@ -1,14 +1,23 @@
...
 		w.WriteHeader(http.StatusOK)
-		_, _ = w.Write([]byte("OK"))
+
+		if _, err := w.Write([]byte("OK")); err != nil {
```
macabu, line 17: Have you had instances of this failing? Trying to understand the motivation behind the PR.

macabu (changes requested):

pkg/api/healthz.go
Alja9: Thanks for the feedback, got it, I'll removed the unnecessary.

pkg/api/healthz.go
Alja9: Got it, I'll removed this header as well and adjust it.
And yes, tools like K8S, docker, and cURL don't cache health check responses, so I'll adjust it.

pkg/api/healthz.go
Alja9: So, basically in my case for observability, when it does fail, we want to know about it via logs and/or traces, instead of silently ignoring failures, we now log and/or record errors on the span, example network timeouts  or broken pipes. 
So, this helps with debugging in PRD when something unexpected happens. It's a small change that adds visibility with minimal overhead and seems worth it for production code, I think.

macabu: I see. Shouldn't the status code also be reflected when it fails, and return a 500?

Alja9: (background: I've running this as pod in k8s)
I think 200 OK is more appropriate because:
1. pod is still healthy, write failure is a transport layer issue not an actual health problem
2. Avoid unnecessary restarts, if returning 500 would case k8s to restart the pod unnecessarily

So my focus now is logging provides visibility with "write fail", error logs and span recording give us the observability.
200 OK + write fail = probe timeout (restart pod)
500 + write fail = probe timeout (restart pod)
In my understanding, both give the same results practically.

macabu: Thanks!

macabu (approved):

leandro-deveikis (approved):

